PR #1, opened 2020-09-13
Add hmf

MatthieuSchaller:
Add the new halo mass function data to the plot summary pages.


JBorrow:
Slightly easier to add plots to this one eh :)

PR #3, opened 2020-09-13
Fix the README

MatthieuSchaller:
Use the correct command to load the submodule and run the data conversion script


JBorrow:
Thanks!

PR #4, opened 2020-09-13
Add baryon- and gas-fractions in haloes

MatthieuSchaller:


MatthieuSchaller:
This needs to be divided by Omega_b / Omega_m on the y-axis to be more useful and hence needs to wait for the cosmology parameters to be read in from the VR catalogues.

MatthieuSchaller:
This is now good to go.

MatthieuSchaller:
Scrap that, we can do much better.

MatthieuSchaller:
Requires https://github.com/SWIFTSIM/velociraptor-python/issues/22 first.

JBorrow:
This should now be possible.

MatthieuSchaller:
All good now. Just need the change to the VR library upstream to fix the look of the x-axis.

gas_fraction_500

JBorrow:
*cries in forgotten $ symbols*

eagle-xl/registration.py
```diff
@@ -164,3 +166,23 @@
         ),
     )
 
+# Now baryon fractions
+
+Omega_m = catalogue.units.cosmology.Om0
+Omega_b = catalogue.units.cosmology.Ob0
+
+M_500 = catalogue.spherical_overdensities.mass_500_rhocrit
```
JBorrow, line 174: This needs to be protected in a similar way as the HI/H2 above.

MatthieuSchaller: Ok, but we don't have a reasonable replacement here. Do you want me to just propose an array of zeros?

JBorrow: I think an array of ones here would be best as that wouldn't lead to crashes. Let's do that for now and I'll revisit this once I have actually got the 'only show this plot if you could make it' working properly.

MatthieuSchaller: Done. :)

MatthieuSchaller:
BTW, this needed a small tweak to the data https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/42

PR #5, opened 2020-09-14
Add plot of the Star Formation Rates function

MatthieuSchaller:


JBorrow (commented):
Is it worth adding these as adaptive? And also could also be helpful to add corresponding histograms.

MatthieuSchaller:
Sure, why not.

I have also reorganised the GSMF file.

JBorrow:
Thanks!

PR #8, opened 2020-09-16
Add HI mass functions to the XL pipeline

MatthieuSchaller:
Does what it says in the title.

### DIFF
--- a/eagle-xl/config.yml
+++ b/eagle-xl/config.yml
@@ -26,6 +26,11 @@ scripts:
     output_file: density_temperature.png
     section: Density-Temperature
     title: Density-Temperature
+  - filename: scripts/subgrid_density_temperature.py
+    caption: Density-temperature diagram constructed from the sub-grid quantities.
+    output_file: density_temperature.png
+    section: Density-Temperature
+    title: Density-Temperature
   - filename: scripts/density_temperature_metals.py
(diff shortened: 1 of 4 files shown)


MatthieuSchaller:
Also adds a subgrid rho-T 2d histogram. 

Slightly annoyingly this one needs more extended axis than the other ones..

JBorrow:
Would it be helpful to also add these to the COLIBRE config?

MatthieuSchaller:
very much so.

JBorrow:
If you would like to add them now you should duplicate them here (make sure you have the correct axis limits that correspond to the change in resolution, though).

MatthieuSchaller:
Let's leave that for the COLIBRists

eagle-xl/config.yml
```diff
@@ -26,6 +26,11 @@ scripts:
     output_file: density_temperature.png
     section: Density-Temperature
     title: Density-Temperature
+  - filename: scripts/subgrid_density_temperature.py
+    caption: Density-temperature diagram constructed from the sub-grid quantities.
+    output_file: subgrid_density_temperature.png
+    section: Density-Temperature
+    title: Density-Temperature
```
JBorrow, line 33: Should be Subgrid Density-Temperature for the title

PR #12, opened 2020-09-17
Add script to plot the birth densities

MatthieuSchaller:


JBorrow:
This plot will be super messy during comparisons; I will come back to you early next week with a better version for the new pipeline.

MatthieuSchaller:
Yeah, thought so. That's where I stopped as the other rho-Z feedback ones are even more messy. I guess for the comparisons we want the same as for the rho-T planes.

MatthieuSchaller:
I guess this can be closer?

JBorrow:
yep thanks

PR #21, opened 2020-10-06
Update colibre config

EvgeniiChaikin:
In this PR I introduce several updates:

- Add stellar birth density distribution plot (compared to EAGLE-XL config, here the y-axis is normalized by the total number of stars to enable comparisons of runs with different particle numbers)
- Add SNII-density distribution plot (the distribution of densities of the particles recorded at the moment when the particles were last heated by SNII)
- Cosmic-time vs Wallclock-time plot (no 2nd y-axis with redshifts yet)
- Number of time-steps vs Wallclock-time plot
- Add more parameters to description.html 
- Correct existing typos in description.html 
- Extend upper limits of the y-axes in the sSFR plots and galaxy-sizes plots


colibre/description.html
```diff
@@ -99,8 +99,19 @@ <h4>Stellar Feedback</h4>
         <td>\(v_{\rm kinetic}\)</td>
         <td>{{ "%.4g" % (data.metadata.parameters["COLIBREFeedback:SNII_delta_v_km_p_s"] | float) }} km s\(^{-1}\)</td>
     </tr>
+     <tr>
+        <td>\(\Delta T\)</td>
+        <td>{{ data.metadata.parameters["COLIBREFeedback:SNII_delta_T_K"] | float | format_number }} K</td>
```
JBorrow, line 104: Please use `.get("COLIBREFeedback:SNII_delta_T_K", default_value)` for things that may not be present in everybody's run.

colibre/description.html
```diff
@@ -99,8 +99,19 @@ <h4>Stellar Feedback</h4>
...
+    </tr>
 </table>
 
+<h4>Gas Cooling</h4>
```
JBorrow, line 108: Not 100% happy with this - what if somebody uses the old Wiersma tables? Is there not something in SWIFT that actually tells us this?

colibre/description.html
```diff
@@ -119,7 +130,20 @@ <h4>Entropy Floor</h4>
...
 
+<h4>Chemistry</h4>
+<table>
+    <tr>
+        <th>Parameter</th>
+        <th>Value</th>
+    </tr>
+    <tr>
+        <td>Metal diffusion constant</td>
+        <td>{{ "%.4g" % (data.metadata.parameters["COLIBREChemistry:metal_diffusion_constant"] | float) }}</td>
```
JBorrow, line 141: use `get()`

colibre/description.html
```diff
@@ -130,27 +154,31 @@ <h4>AGN</h4>
...
+        <td>BH Subgrid Seed Mass</td>
         <td>{{ data.metadata.parameters["COLIBREAGN:subgrid_seed_mass_Msun"] | float | format_number }} M\(_\odot\)</td>
     </tr>
     <tr>
-        <td>N\(_{\rm heat}\)</td>
+        <td>Seed Halo Mass</td>
+        <td>{{ data.metadata.parameters["FOF:black_hole_seed_halo_mass_Msun"] | float | format_number }} M\(_\odot\)</td>
+    </tr>
+    <tr>
+        <td>Targer number of particles to heat N\(_{\rm heat}\)</td>
```
JBorrow, line 173: Target

colibre/description.html
```diff
@@ -159,29 +187,60 @@ <h4>AGN</h4>
...
+        <td>{{ data.metadata.parameters.get("COLIBREAGN:with_krumholz_vorticity", 0) | int }}</td>
     </tr>
     <tr>
         <td>Use Angular Momentum Limiter</td>
         <td>{{ data.metadata.parameters.get("COLIBREAGN:with_angmom_limiter", 0) | int }}</td>
     </tr>
     <tr>
-        <td>Reposition Velocity</td>
-        <td>{{ data.metadata.parameters.get("COLIBREAGN:reposition_coefficient_upsilon", 0.0) | float | format_number }}</td>
+        <td>Normalisation of the viscous time-scale</td>
```
JBorrow, line 197: Can we keep this as `viscous_alpha` please? That's more clear to me.

colibre/description.html
```diff
@@ -159,29 +187,60 @@ <h4>AGN</h4>
...
+        <td>Repositioning Speed Normalisation</td>
+	<td>{{ data.metadata.parameters.get("COLIBREAGN:reposition_coefficient_upsilon", 0.0) | float | format_number }} km s\(^{-1}\) M\(_\odot\)\(^{-1}\)</td>
+    </tr>
+    <tr>
+        <td>Black Hole Merger Threshold Type</td>
+        <td>{{ data.metadata.parameters.get("COLIBREAGN:merger_threshold_type", 0) | int }}</td>
+    </tr>
+    <tr>
+        <td>Black Hole Merger Max Distance Ratio</td>
+        <td>{{ "%.4g" % (data.metadata.parameters["COLIBREAGN:merger_max_distance_ratio"] | float) }}</td>
```
JBorrow, line 238: Use `get`. Also a lot of these names are quite long - do they appear in the table ok?

colibre/scripts/SNII_density_distribution.py
```diff
@@ -0,0 +1,117 @@
...
+from swiftsimio import load
+
+from unyt import mh, cm, Gyr
+from matplotlib.colors import LogNorm, Normalize
+from matplotlib.animation import FuncAnimation
+
+from swiftpipeline.argumentparser import ScriptArgumentParser
+
+arguments = ScriptArgumentParser(
+    description="Creates a star formation history plot, with added observational data."
```
JBorrow, line 18: No, it does not :)

colibre/scripts/birth_density_distribution.py
```diff
@@ -0,0 +1,104 @@
...
+from swiftsimio import load
+
+from unyt import mh, cm, Gyr
+from matplotlib.colors import LogNorm, Normalize
+from matplotlib.animation import FuncAnimation
+
+from swiftpipeline.argumentparser import ScriptArgumentParser
+
+arguments = ScriptArgumentParser(
+    description="Creates a star formation history plot, with added observational data."
```
JBorrow, line 18: Again, this needs to be changed.

colibre/scripts/wallclock_simulation_time.py
```diff
@@ -0,0 +1,63 @@
...
+run_names = arguments.name_list
+run_directories = [f"{directory}" for directory in arguments.directory_list]
+snapshot_names = [f"{snapshot}" for snapshot in arguments.snapshot_list]
+output_path = arguments.output_directory
+
+plt.style.use(arguments.stylesheet_location)
+
+fig, ax = plt.subplots()
+
+for run_name, run_directory, snapshot_name in zip(run_names,
```
JBorrow, line 33: Please format all of your scripts with `black`.

JBorrow (commented):
Phew! A lot of changes here, thanks. Please can you make sure everything is formatted with `black`, and in the future it would be nicer to have smaller self-contained PRs. Thanks for the contribution!

colibre/description.html
EvgeniiChaikin: That is in principle true for all stellar-feedback related fields. Shall I pick `0.0` as the default value for all of them?

JBorrow: Yes, that would be best, especially at this stage where we might add/remove fields.

colibre/description.html
EvgeniiChaikin: Personally, I'd be very surprised if we decide to use something different alongside the COLIBRE pipeline (in addition to the mentioned two cooling models). 

Nonetheless, I can change `Table cooling (Ploeckinger, 2020)` -> `Table cooling` to make it more general, if you think that'd be better.

colibre/description.html
EvgeniiChaikin: Sure

colibre/description.html
JBorrow: Ah my point was more that surely there's something in the SWIFT output that tells us what cooling model we're using? There is for hydro, gravity, etc.

colibre/description.html
EvgeniiChaikin: Alright, 

Also, I've been thinking, sometimes it might be unclear if a parameter is not specified or explicitly set to zero. Maybe in the former case we could write `None` or something like that.

colibre/description.html
EvgeniiChaikin: Screenshot from 2020-10-07 17-28-52

The table looks fine. Well, okay, maybe there is a bit too much text. What do you think?

colibre/description.html
EvgeniiChaikin: I am not sure if there is anything in the hydro part of a snapshot that would tell us something about the cooling.

The easiest thing that comes to my mind is to look for a whole block of paramteres `COLIBRECooling:parameters` and `CHIMESCooling:parameters` .

In order to cover all possible cases, my suggestion would be to

- look for something from `COLIBRECooling:parameters`; if we find anything then we are using COLIBRE cooling,
- if not, then look for something from `CHIMESCooling:parameters`; if we get something then we know that CHIMES cooling was adopted
- Say `Unknown`

JBorrow: Right but we should probably change this in SWIFT then. We have sections in `SubgridScheme` that should tell us about this.

colibre/scripts/wallclock_simulation_time.py
EvgeniiChaikin: Done

colibre/scripts/birth_density_distribution.py
EvgeniiChaikin: Corrected

colibre/scripts/SNII_density_distribution.py
EvgeniiChaikin: Thanks, now it does

colibre/description.html
JBorrow: So I've just checked and

`data.metadata.subgrid_scheme["Cooling Model"].decode("utf-8")`

Should give us the cooling model. If it does not for COLIBRE, we need to fix it.

colibre/description.html
JBorrow: Indeed but this is more complex when you pass everything through the float and format_number thing... I could write a function called get_if_present_float and get_if_present_int or something? I think the best idea if not present is just to have an empty string.

EvgeniiChaikin: Yes, that'd be great.

Sure, let's use an empty string.

colibre/description.html
EvgeniiChaikin: Indeed! It works
Screenshot from 2020-10-09 12-01-21

EvgeniiChaikin:
Anything else that needs to be improved in this PR?

JBorrow:
Let's wait until I have been able to produce the functions to deal with possibly non-existent properties before merging this.

JBorrow:
This can now be done, please update your branch. See the `eaglexl/description.html` for usage. There is also an integer version of the function.

EvgeniiChaikin:
I incorporated `get_if_present_int()` and `get_if_present_float()` into `colibre/description.html`

colibre/scripts/SNII_density_distribution.py
```diff
@@ -0,0 +1,127 @@
...
+
+try:
+    plt.style.use("mnras.mplstyle")
+except:
+    pass
+
+from swiftpipeline.argumentparser import ScriptArgumentParser
+
+arguments = ScriptArgumentParser(
+    description="Creates a plot showing the distribution of the gas densities recorded when the gas was last heated by SNII, split by redshift"
```
JBorrow, line 21: This is quite long, can you split it up as:

```suggestion
    description=(
        "Creates a plot showing the distribution of the gas densities "
        "recorded when the gas was last heated by SNII, split by redshift"
    )
```

JBorrow:
After the one minor change I am happy to merge.

colibre/scripts/SNII_density_distribution.py
EvgeniiChaikin: Should be done now. There are also a few other places in the code where the limit of 88 characters per line is exceeded. I can change these as well if needed.

JBorrow: the rest are fine if they are just strings, this one was just particularly egregious.

PR #24, opened 2020-10-10
Metallicity evolution plots

MatthieuSchaller:
Extending the fun a bit more

### DIFF
--- a/eagle-xl/scripts/bh_metallicity_evolution.py
+++ b/eagle-xl/scripts/bh_metallicity_evolution.py
@@ -1,0 +1,93 @@
+"""
+Plots the stellar metallicity density evolution.
+"""
+import matplotlib
+
+import unyt
+
+import matplotlib.pyplot as plt
+import numpy as np
+import sys
+
+
+from swiftsimio import load, load_statistics
+
+from swiftpipeline.argumentparser import ScriptArgumentParser
+
+arguments = ScriptArgumentParser(
+    description="Creates a metallicity density evolution plot for gas, stars and black holes."
+)
+
+snapshot_filenames = [
+    f"{directory}/{snapshot}"
+    for directory, snapshot in zip(arguments.directory_list, arguments.snapshot_list)
+]
+
+stats_filenames = [
+    f"{directory}/statistics.txt" for directory in arguments.directory_list
+]
+
+names = arguments.name_list
+output_path = arguments.output_directory
+
+plt.style.use(arguments.stylesheet_location)
+
+simulation_lines = []
+simulation_labels = []
+
+fig, ax = plt.subplots()
+
+ax.loglog()
+
+for snapshot_filename, stats_filename, name in zip(
+    snapshot_filenames, stats_filenames, names
+):
+    data = load_statistics(stats_filename)
+
+    snapshot = load(snapshot_filename)
+    boxsize = snapshot.metadata.boxsize
+    box_volume = boxsize[0] * boxsize[1] * boxsize[2]
+
+    # a, Redshift, SFR
+    scale_factor = data.a
+    redshift = data.z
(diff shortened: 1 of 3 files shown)


eagle-xl/scripts/bh_metallicity_evolution.py
```diff
@@ -0,0 +1,93 @@
...
+    redshift = data.z
+    bh_Z_mass = data.bh_z_mass
+    bh_Z_mass_density = bh_Z_mass / box_volume
+
+    # High z-order as we always want these to be on top of the observations
+    simulation_lines.append(ax.plot(scale_factor, bh_Z_mass_density, zorder=10000)[0])
+    simulation_labels.append(name)
+
+ax.set_xlabel("Redshift $z$")
+ax.set_ylabel(r"Metal Mass in Black Holes $\rho_{\rm BH,Z}$ [M$_\odot$ Mpc$^{-3}$]")
```
JBorrow, line 62: This is not necessarily true. Please use `bh_Z_mass_density.to("Solar_Mass / Mpc**3")`.

JBorrow: In fact it won't be - the value produced will be 1e10 M_\odot Mpc^-3

eagle-xl/scripts/metallicity_evolution.py
```diff
@@ -0,0 +1,124 @@
...
+            star_Z_mass_density,
+            linestyle="dashed",
+            color=f"C{color}",
+            zorder=10000,
+        )[0]
+    )
+    simulation_labels.append(name)
+
+ax.set_xlabel("Redshift $z$")
+ax.set_ylabel(r"Metal Mass $\rho_{\rm Z}$ [M$_\odot$ Mpc$^{-3}$]")
```
JBorrow, line 81: Again requires use of the unit conversions. Alternatively, you can use the `unyt.matplotlib_support` environment (see https://unyt.readthedocs.io/en/stable/usage.html#plotting-with-matplotlib)

JBorrow (commented):
Needs to use the unit conversions. Are these missing from the existing plots too?

MatthieuSchaller:
I thought your reading script was handling the unit conversions correctly. The numbers that come out make sense currently, so I don't think anything needs changing here.

JBorrow:
What I mean is that it's by chance. They are symbolic units - they may be in CGS, depending on the output. If you force a conversion then we _know_ it is in those units. At the moment I guess it's by chance correct but that may not always be the case.

MatthieuSchaller:
Ok but then I don't understand what you want me to do here.

eagle-xl/scripts/bh_metallicity_evolution.py
```diff
@@ -0,0 +1,93 @@
...
+
+    snapshot = load(snapshot_filename)
+    boxsize = snapshot.metadata.boxsize
+    box_volume = boxsize[0] * boxsize[1] * boxsize[2]
+
+    # a, Redshift, SFR
+    scale_factor = data.a
+    redshift = data.z
+    bh_Z_mass = data.bh_z_mass
+    bh_Z_mass_density = bh_Z_mass / box_volume
```
JBorrow, line 55: If here you do `bh_Z_mass_density = (bh_Z_mass / box_volume).to("Solar_Mass / Mpc**3")` you will always ensure that the value that you plot has those units, if that makes sense.

JBorrow:
Great, thanks. I'll take another pass at this tomorrow before merging but it looks good to me.

One case where this would have definitely gone wrong is simulations performed with kpc as the internal length unit. Not that we're doing that right now, but you never know...

MatthieuSchaller:
Yes, makes sense now.

JBorrow:
Thanks!

PR #27, opened 2020-10-14
Adaptive Bins for XL

JBorrow:


EvgeniiChaikin:
@JBorrow `eagle-xl/scripts/__pycache__/load_sfh_data.cpython-38.pyc` should not have been part of the PR, right?

JBorrow:
Whoops, indeed that should not be here. I'll fix that as soon as I can.

PR #28, opened 2020-10-14
Add plots of the BH evolution through time

MatthieuSchaller:
Two new plots:
 - BH accretion rate through cosmic time
 - BH mass through cosmic time.

### DIFF
--- a/eagle-xl/scripts/bh_accretion_evolution.py
+++ b/eagle-xl/scripts/bh_accretion_evolution.py
@@ -1,0 +1,124 @@
+"""
+Plots the BH accretion rate density evolution
+"""
+import matplotlib
+
+import unyt
+
+import matplotlib.pyplot as plt
+import numpy as np
+import sys
+import glob
+
+
+from swiftsimio import load, load_statistics
+
+from swiftpipeline.argumentparser import ScriptArgumentParser
+
+from velociraptor.observations import load_observation
+
+arguments = ScriptArgumentParser(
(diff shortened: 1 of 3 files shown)


MatthieuSchaller:
Actually, there might be a funky bug here. If you have a section appearing both in the auto_plotter and the config then you get two sections with the same name (and a broken html anchor).

JBorrow:
Thanks, I've opened up an issue for that. I don't want to address that for this PR though.

eagle-xl/scripts/bh_accretion_evolution.py
```diff
@@ -0,0 +1,124 @@
...
+for index, observation in enumerate(observational_data):
+    obs = load_observation(observation)
+    obs.plot_on_axes(ax)
+
+ax.set_xlabel("Redshift $z$")
+ax.set_ylabel(r"BH Accretion Rate Density [M$_\odot$ yr$^{-1}$ Mpc$^{-3}$]")
+
+redshift_ticks = np.array([0.0, 0.2, 0.5, 1.0, 2.0, 3.0, 5.0, 10.0, 20.0, 50.0, 100.0])
+redshift_labels = [
+    "$0$",
```
JBorrow, line 93: Fwiw these probably don't need to be dollar signed

JBorrow:
Looks good to me. If it makes your life easier remember you can use `{item}.units.latex_repr` to get a pre-written LaTeX representation of a symbolic unit.

PR #31, opened 2020-10-20
Mass plots

MatthieuSchaller:
Add two plots to help see what is going in the simulations:
 - Simple histogram of all the gas particle masses
 - Scatter plot of the BH subgrid and dynamical masses

### DIFF
--- a/eagle-xl/scripts/bh_masses.py
+++ b/eagle-xl/scripts/bh_masses.py
@@ -1,0 +1,116 @@
+"""
+Makes a BH dyn. mass vs. subgrid mass plot. Uses the swiftsimio library.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from swiftsimio import load
+
+from unyt import unyt_quantity
+from matplotlib.colors import LogNorm
+from matplotlib.animation import FuncAnimation
+
+mass_bounds = [6e3, 6e9]
+
+
+def get_data(filename):
+    """
+    Grabs the data (masses in Msun).
+    """
+
+    data = load(filename)
+
+    mass_sub = data.black_holes.subgrid_masses.to("Msun")
+    mass_dyn = data.black_holes.dynamical_masses.to("Msun")
+    mass_seed = unyt_quantity(
+        float(data.metadata.parameters["EAGLEAGN:subgrid_seed_mass_Msun"]), "Msun"
+    )
+    mass_gas = np.min(data.gas.masses.to("Msun"))
+
+    return mass_sub, mass_dyn, mass_seed, mass_gas
+
+
+def setup_axes(mass_bounds, number_of_simulations: int):
+    """
+    Creates the figure and axis object. Creates a grid of a x b subplots
+    that add up to at least number_of_simulations.
+    """
+
+    sqrt_number_of_simulations = np.sqrt(number_of_simulations)
+    horizontal_number = int(np.ceil(sqrt_number_of_simulations))
+    # Ensure >= number_of_simulations plots in a grid
+    vertical_number = int(np.ceil(number_of_simulations / horizontal_number))
+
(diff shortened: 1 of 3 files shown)


eagle-xl/scripts/bh_masses.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Grabs the data (masses in Msun).
+    """
+
+    data = load(filename)
+
+    mass_sub = data.black_holes.subgrid_masses.to("Msun")
+    mass_dyn = data.black_holes.dynamical_masses.to("Msun")
+    mass_seed = unyt_quantity(
+        float(data.metadata.parameters["EAGLEAGN:subgrid_seed_mass_Msun"]), "Msun"
```
JBorrow, line 27:
```suggestion
        float(data.metadata.parameters.get("EAGLEAGN:subgrid_seed_mass_Msun", 0.0)), "Msun"
```
will not break in the case where this script is accidentally ran on a different sub-grid model.

eagle-xl/scripts/bh_masses.py
```diff
@@ -0,0 +1,116 @@
...
+
+    ax = np.array([ax]) if number_of_simulations == 1 else ax
+
+    if horizontal_number * vertical_number > number_of_simulations:
+        for axis in ax.flat[number_of_simulations:]:
+            axis.axis("off")
+
+    # Set all valid on bottom row to have the horizontal axis label.
+    for axis in np.atleast_2d(ax)[:][-1]:
+        axis.set_xlabel("Black Hole Dynamical Masses $M_{\\rm dyn}$ [M$_\odot$]")
```
JBorrow, line 57: These are quite long and will overlap significantly in the case where we have multiple simulations plotted. Can you do something like "BH Dyn. Masses $M_{\\rm dyn}$ [M$_\odot$]"?

eagle-xl/scripts/gas_masses.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def get_data(filename):
+    """
+    Grabs the data (masses in 10**6 Msun).
+    """
+
+    data = load(filename)
+
+    mass_gas = data.gas.masses.to("Msun") / 10 ** 6
```
JBorrow, line 24: I think it's better practice here to do `data.gas.masses.to("1e6 * Msun")`

eagle-xl/scripts/gas_masses.py
```diff
@@ -0,0 +1,108 @@
...
+    mass_split = unyt_quantity(
+        1e4 * float(data.metadata.parameters["SPH:particle_splitting_mass_threshold"]),
```
JBorrow, line 26: Same comment as above about the use of `get()`; also why is this 1e4 here?

JBorrow (commented):
Just a few minor comments.

eagle-xl/scripts/gas_masses.py
MatthieuSchaller: Because this is expressed in internal units. so 10^-4 * 10^10 = 10^6.

I could first convert the gas masses to 10^10 Msun and then multiply by 10^4 for converting to 10^6 Msun.

JBorrow: Ah - can we then do:
```
mass_split = unyt_quantity(
    data.metadata.parameters.get("SPH:particle_splitting_mass_threshold", 0.0),
    units=data.units.mass,
).to("1e6 * Msun")
```

JBorrow: Just trying to cover off a different set of mass units being chosen - however unlikely.

MatthieuSchaller:
Done.

JBorrow:
Great thanks!

PR #34, opened 2020-10-21
Two tiny (stylistic) fixes

EvgeniiChaikin:
- **First**
Screenshot from 2020-10-21 00-39-46

Since we are now using `get_if_present_float() / get_if_present_int()`, which add units to the parameters we pass to the functions  _by default_, specifying **kpc** next to **baryons** & **dark matter** in the image above is redundant.

- **Second**

Screenshot from 2020-10-21 00-39-23

In certain cases,
```
try:	
    plt.style.use("mnras.mplstyle")	
except:	
    pass
```
seems to give undesired padding to the figures (_left panel of the figure above_). Removing those 5 lines of code shrinks the panning layer to zero  (_right panel of the figure above_), which is what we want.

### DIFF
--- a/colibre/scripts/SNII_density_distribution.py
+++ b/colibre/scripts/SNII_density_distribution.py
@@ -9,12 +9,6 @@ import unyt
 from unyt import mh, cm
 
 from swiftsimio import load
-
-try:
-    plt.style.use("mnras.mplstyle")
-except:
-    pass
-
 from swiftpipeline.argumentparser import ScriptArgumentParser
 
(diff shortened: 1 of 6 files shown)


colibre/scripts/SNII_density_distribution.py
```diff
@@ -9,12 +9,6 @@
 from unyt import mh, cm
 
 from swiftsimio import load
-
```
JBorrow, line 12: Indeed this is very odd and should not have been there! Thanks for catching.

PR #38, opened 2020-10-22
Changed gas_masses and max_temperatures to promote better comparisons…

JBorrow:
… between runs

Before even though we were only plotting a single line, each run was plotted on a single sub-figure. I don't think this is the best way of showing off the data. See the below for the 'improved' figures that use one sub-figure but multiple coloured lines.

gas_masses

gas_max_temperatures


MatthieuSchaller:
Yes, good idea. Though I'd keep the linear x-axis for the masses given the narrow range of values we use.

PR #40, opened 2020-10-22
Correct the Mstar - Mhalo relation plot to use Mvir and the Behroozi+…

MatthieuSchaller:
…2019 data


MatthieuSchaller:
http://icc.dur.ac.uk/~jlvc76/SWIFT/output_master_oct17/#2049692207179438550

JBorrow:
Can we call them M_BN98 instead of M_vir? I think that M_vir is a bit ambigious.

JBorrow:
Thanks!

PR #41, opened 2020-10-24
Add redshift axis to run-performance plots

EvgeniiChaikin:
Add redshift ticks to run-performance plots. The ticks will be attached to the second Y (or X) axis

To do the conversion from z to cosmic time I use **Planck13** cosmology, which is available in `astropy.cosmology`

Planck13 cosmology is the cosmology adopted in EAGLE and currently used in COLIBRE.

**Examples:**

_Before the change_

simulation_time_number_of_steps_before
wallclock_simulation_time_before

_After the change_

simulation_time_number_of_steps
wallclock_simulation_time

### DIFF
--- a/colibre/scripts/SNII_density_distribution.py
+++ b/colibre/scripts/SNII_density_distribution.py
@@ -13,7 +13,7 @@ from swiftpipeline.argumentparser import ScriptArgumentParser
 
 arguments = ScriptArgumentParser(
     description="Creates a plot showing the distribution of the gas densities recorded "
-                "when the gas was last heated by SNII, split by redshift"
+    "when the gas was last heated by SNII, split by redshift"
 )
 
 snapshot_filenames = [
--- a/colibre/scripts/simulation_time_number_of_steps.py
+++ b/colibre/scripts/simulation_time_number_of_steps.py
@@ -27,8 +27,8 @@ fig, ax = plt.subplots()
 # We will need to keep track of the maximum cosmic time reached in the simulation(s)
 t_max = unyt.unyt_array(0.0, units="Gyr")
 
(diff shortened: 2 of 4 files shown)


EvgeniiChaikin:
I also updated the star-formation history plot. I will look like this
star_formation_history

JBorrow:
We have discussed this before - please use:

```
data = load("snapshot_xxxx.hdf5")
cosmology = data.metadata.cosmology
```

EvgeniiChaikin:
Yes, we did. And as far as remember, we agreed that the best way of doing this would be to use an astropy object and a popular cosmology, like `Planck13` (which is the cosmology currently adopted in COLIBRE).

```
data = load("snapshot_xxxx.hdf5")
cosmology = data.metadata.cosmology
```
does not seem to be a perfect solution to me because if we have several runs' data in the plot, then we will always have to choose the cosmological parameters from one particular run and use these for the other runs.

Also, importing `Planck13` (rather than `cosmology = data.metadata.cosmology`) makes it very explicit in the script itself what cosmological model is used.

JBorrow:
Right, but your claim that if the runs cosmology vary then we can't choose a correct one is not consistent with choosing a possibly always-wrong one, surely? I think we can assume that the cosmology is always the same, and just choose the first snapshot to extract the cosmology from. I am very wary of baking in things like this, as they always inevitably lead to errors.

EvgeniiChaikin:
I agree with what you are saying, but at the same time (at least to me) it is difficult to imagine a universe wherein Planck13 is considered to be a wrong cosmology :)

I updated the PR. Now it uses the cosmological model from the first run fed to the pipeline.

JBorrow:
There are actually many reasons to choose a cosmology that is 'wrong' :).

PR #42, opened 2020-10-24
Add M_BN98 SMHM plots to colibre config

EvgeniiChaikin:
Add M_BN98 stellar-mass-halo-mass plots to colibre config including `Berhoozi2019 data`

**Example:**

Screenshot from 2020-10-24 19-28-45

### DIFF
--- a/colibre/auto_plotter/stellar_mass_halo_mass.yml
+++ b/colibre/auto_plotter/stellar_mass_halo_mass.yml
@@ -28,7 +28,6 @@ stellar_mass_halo_mass_M200_all_100:
     section: Stellar Mass-Halo Mass
   observational_data:
     - filename: GalaxyStellarMassHaloMass/Moster2013.hdf5
-    - filename: GalaxyStellarMassHaloMass/Behroozi2013.hdf5
 
 stellar_mass_halo_mass_MBN98_all_100:
   type: "scatter"
@@ -94,7 +93,6 @@ stellar_mass_halo_mass_M200_centrals_100:
     section: Stellar Mass-Halo Mass
   observational_data:
     - filename: GalaxyStellarMassHaloMass/Moster2013.hdf5
-    - filename: GalaxyStellarMassHaloMass/Behroozi2013.hdf5
     - filename: GalaxyStellarMassHaloMass/Schaye2015_Ref_100_z0p0.hdf5
 
 stellar_mass_halo_mass_MBN98_centrals_100:
@@ -162,7 +160,6 @@ stellar_mass_halo_mass_M200_centrals_ratio_100:
     section: Stellar Mass-Halo Mass
   observational_data:
     - filename: GalaxyStellarMassHaloMass/Moster2013Ratio.hdf5
-    - filename: GalaxyStellarMassHaloMass/Behroozi2013Ratio.hdf5
     - filename: GalaxyStellarMassHaloMass/Schaye2015_Ref_100_z0p0_Ratio.hdf5
 
 stellar_mass_halo_mass_MBN98_centrals_ratio_100:
@@ -230,7 +227,6 @@ stellar_mass_halo_mass_M200_centrals_ratio_stellar_100:
     section: Stellar Mass-Halo Mass
   observational_data:
     - filename: GalaxyStellarMassHaloMass/Moster2013RatioStellar.hdf5
-    - filename: GalaxyStellarMassHaloMass/Behroozi2013RatioStellar.hdf5
     - filename: GalaxyStellarMassHaloMass/Schaye2015_Ref_100_z0p0_RatioStellar.hdf5
       
 stellar_mass_halo_mass_MBN98_centrals_ratio_stellar_100:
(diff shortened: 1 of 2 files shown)


JBorrow:
Do you want to show all of these on the webpage or would a subset suffice?

EvgeniiChaikin:
I am fine with either. I chose 4 plots of MBN98 because we currently have 4 plots of M200crit. In this way, the number of MBN98 and M200crit plots is the same, and hence everything looks symmetric. @JBorrow @MatthieuSchaller what do you think?

EvgeniiChaikin:
@MatthieuSchaller Also I think we could move `Berhoozi 2013` from M200crit to MBN98 plots. Then the correction factor of **1.2** for the mass (which is only valid for `z=0`) is no longer needed.

JBorrow:
You'll need to rebase

JBorrow:
Imo it's too many plots - i would suggest maybe cutting the non centrals-only ones from the webpage, and maybe the un-reduced M_BN98 one.

MatthieuSchaller:
In the XL pipeline, I am now showing this:

 - Mvir - M* (all) with B19 data
 - Mvir - M* (centrals) with B19 data
 - M200 - M* (all) with M13 data and EAGLE data
 - M200 - M* (centrals) with M13 data and EAGLE data
 - M200 - M*/M200 (centrals) with M13 data and EAGLE data
 - M* - M*/M200 (centrals) with M13 data and EAGLE data

I think this is enough to cover all we need. Maybe we can also add the M19 data instead of M13.

Note that I have not forgotten to produce the EAGLE data for M_vir and z>0. It's getting there :)

EvgeniiChaikin:
Don't we also want to show M*/MBN98 vs MBN98? (or in your notation M*/Mvir vs Mvir) I thought that was the most crucial plot.

MatthieuSchaller:
You can but then I'd remove the M_200 version. Too many plots just aren't useful.

EvgeniiChaikin:
I presume 6 plots should be fine.

 I am not going to display the `M_BN98` plot for all galaxies (i.e. including satellites) because Berhoozi2019 data actually gives `M_BN98,max` not `M_BN98`. These two might be very different for satellites.

And also I think showing M*/M_BN98 vs stellar mass is not necessary.

EvgeniiChaikin:
**A new example** 
Screenshot from 2020-10-26 11-47-02

JBorrow:
I don't think your M*/M_H vs M_H plots are actually only showing centrals.

EvgeniiChaikin:
What do you mean exactly? For every plot with the  `comment: "Centrals only" ` I also write `select_structure_type: 10` in the .yml file.

EvgeniiChaikin:
Indeed, MBN98 plots have more objects than they should...

JBorrow:
I mean more that in your plots (for both halo definitions of M_* / M_H against halo mass, there are a lot of low mass haloes with high stellar fractions that aren't present in the other plots. I don't know why that is though :)

EvgeniiChaikin:
The .yml file seems to be fine. Perhaps the cut `select_structure_type: 10` does not properly work when applied to `halo_mass_bn98`?

JBorrow:
Hmm, well it definitely _should_. I don't know why it's not working though. Or maybe my understanding of the M_* / M_H relation is wrong. I guess we see this in the other suite too so it's probably not a problem.

JBorrow:
happy to merge when you're ready.

EvgeniiChaikin:
If you look at the plot in the description of this MR, there is indeed no difference between the MBN98 plots with (second panel) and without (fourth panel) satellites. So something is not working right...

EvgeniiChaikin:
The description was **outdated.** That is what confused me. Everything actually works correctly.

**Example with satellites**

stellar_mass_halo_mass_MBN98_all_ratio_100

Example without satellites

stellar_mass_halo_mass_MBN98_centrals_ratio_100

**Example of how the plots will appear on the html page:**

Screenshot from 2020-10-26 16-58-01

So I think this PR can be merged.

MatthieuSchaller:
Thanks for chasing this down. Maybe one last suggestion: removing the B13 "data" from the plots since it is not correct at z>0 and we have the better version.

EvgeniiChaikin:
The B13 are also produced using the fitting function (hence we can generate these at various z). The problem with the mass conversion factor can be tackled by moving the B13 data to M_BN98 plots. Or do you think it is best to get rid of B13 completely?

MatthieuSchaller:
I'd get rid of the conversion factor altogether that was a dirty quick fix. Either move it to the new plots or get rid of it altogether.

EvgeniiChaikin:
OK, I will make that change. But I would rather do it in a separate PR.

EvgeniiChaikin:
In this one, I can delete the B13 data from .yml files if that's what we want.

MatthieuSchaller:
No need to get rid of the data files just not plotting them.

EvgeniiChaikin:
Done!

Looks a bit empty without B13 data thought

PR #44, opened 2020-10-24
Make all histograms cumulative

EvgeniiChaikin:
Make all histograms in colibre config **cumulative** (the cumulative sum is from _high_ to _low_)

**Example** of how it will look like on html webpage:

_Before the change_

Screenshot from 2020-10-24 20-16-29

_After the change_

Screenshot from 2020-10-24 20-16-17


JBorrow:
Looks good, but it might be good to retain the non-cumulative histograms (not showing them on the webpage) as this is important for if we come to analyse the data later.

EvgeniiChaikin:
Indeed! It is best to have both.

EvgeniiChaikin:
This change is now in this PR

PR #45, opened 2020-10-24
Add 6 scripts from eagleXL

EvgeniiChaikin:
**Copy 6 scripts from eagleXL and add them to colibre**

- BH dynamical vs subgrid mass
- BH mass density versus redshift
- BH accretion rate versus redshift
- Gas/stellar metallicity versus redshift
- BH metallicity versus redshift
- Gas-particle mass historgram

### DIFF
--- a/colibre/scripts/metallicity_evolution.py
+++ b/colibre/scripts/metallicity_evolution.py
@@ -66,15 +66,16 @@ for color, (snapshot_filename, stats_filename, name) in enumerate(
             zorder=10000,
         )[0]
     )
-    simulation_lines.append(
-        ax.plot(
-            scale_factor,
-            star_Z_mass_density,
-            linestyle="dashed",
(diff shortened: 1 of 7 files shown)


colibre/scripts/metallicity_evolution.py
```diff
@@ -0,0 +1,124 @@
...
+    simulation_lines.append(
+        ax.plot(
+            scale_factor,
+            gas_Z_mass_density,
+            linestyle="solid",
+            color=f"C{color}",
+            zorder=10000,
+        )[0]
+    )
+    simulation_lines.append(
```
JBorrow, line 69: this shouldn't be appended here - we just fixed this bug on the xl side.

EvgeniiChaikin: Fixed

PR #46, opened 2020-10-26
New stellar birth density plots

EvgeniiChaikin:
This one is a pretty big one.

In this update, I introduce

- two new scripts (stellar birth density vs. redshift and stellar birth density vs. metallicity). In order to create these two plots, I merely took the template from the temperature-density diagram script. 
- Max & average stellar birth density as a function of stellar mass and as a function of halo mass, for each halo. To create these plots, in addition to this pull request, I created a stellar-birth-density registration function [HERE](https://github.com/SWIFTSIM/velociraptor-python/pull/32) and update the velociraptor config as shown [HERE](https://gitlab.cosma.dur.ac.uk/EAGLE/swift-colibre/-/merge_requests/143).

**Examples** This is how the stellar-birth-density section is expected to look like following this update

Screenshot from 2020-10-26 21-08-26

and in the **comparison** mode

Screenshot from 2020-10-26 21-55-43

@JBorrow @MatthieuSchaller

### DIFF
--- a/colibre/auto_plotter/stellar_birth_densities.yml
+++ b/colibre/auto_plotter/stellar_birth_densities.yml
@@ -9,8 +9,8 @@ stellar_birth_density_halo_mass_average_all_100:
   y:
     quantity: "stellar_birth_densities.average"
     units: "mh/cm**3"
-    start: 1e-1
-    end: 1e6
+    start: 5e0
+    end: 5e5
   median:
     plot: true
     log: true
@@ -38,8 +38,8 @@ stellar_birth_density_halo_mass_max_all_100:
   y:
     quantity: "stellar_birth_densities.max"
     units: "mh/cm**3"
-    start: 1e-1
-    end: 1e6
+    start: 5e0
+    end: 5e5
   median:
     plot: true
     log: true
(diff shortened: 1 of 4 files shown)


EvgeniiChaikin:
Btw, it looks like most people have runs' names that are usually too long. These long names make temperature-density and similar diagrams undergo weird behaviour and shrink. How about we change the run names to what is shown below?

Screenshot from 2020-10-26 21-22-03

I discussed it with @JBorrow a month ago or so. Back then my suggestion was to enable the pipeline to accept arbitrary names entered by the user. We agreed that it is not best practice because it might lead to mistakes. But how about we use a very simple name convetion (Run1, Run2, Run3, etc), as demonstrated in the plot above?

EvgeniiChaikin:
Here's an example of a whole html page
 [LINK](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/With_AGN/L006N188/With_BH_mergers/z2.0_FID_Oct2020/)

JBorrow:
I don't agree that naming things Run{1,2,3} is good practice. I think we will in the end probably have to revisit allowing custom run names, despite my best protests...

JBorrow:
Can I ask why you always have the box size and particle number in the run name? This information is in the snapshot anyway.

EvgeniiChaikin:
There are mainly two reasons:

1) Because I ofter vary the box size or/and the particle number
2) I always what to know the box/resolution each line in the plot refers to

PR #49, opened 2020-10-27
Small improvements and fixes to colibre cfg

EvgeniiChaikin:
**A bunch of small improvements and fixes to colibre config.**

- Extend the stellar-mass axis down to 5e4 solar mass in all SMHM plots 
- Fix the bug (it should be logaverage not (linear) average)
- Add basic info on the dust model to html page.


EvgeniiChaikin:
I received a request to plot the stellar mass down to 9e3 in place of 5e4

PR #50, opened 2020-10-27
Adding stellar FeH plot

correac:
Hi Josh! This is a very WiP attempt. I would like to add a few plots to the pipeline, such as log mean(Fe/H) abundances or mean of log, and also some plots for metal distribution functions.

When you have time, could you have a look and send me comments?

On a side note : I have trouble in line 66 of stellar_mass_mean_stellar_FeH.py

Btw I have been looking at swiftsimio, xl-pipeline and velociraptor, they look super!

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -56,11 +56,15 @@ scripts:
     output_file: metallicity_distribution.png
     section: Metal Mass Fractions
     title: Metal Mass Fraction Distribution
-  - filename: scripts/stellar_mass_mean_stellar_FeH.py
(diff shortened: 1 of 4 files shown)


EvgeniiChaikin:
@JBorrow looks like this PR has been here for a while. If you don't have time, I can have a look

JBorrow:
Apologies for taking so long to get to this. @correac, did we decide that this was quite an expensive script? It seems so.

EvgeniiChaikin:
I am closing this branch because Fe/H plots will be added in #114

PR #51, opened 2020-10-27
allow registration of species for table-cooling runs, and fix error printing in dust name

wmfw23:
A couple of fixes following testing


JBorrow:
Hi James can you check that I fixed the conflicts correctly?

PR #57, opened 2020-11-06
Add H_2 mass function data

MatthieuSchaller:


JBorrow:
Can you also rev the commit of the obesrvational data if this is newly added?

MatthieuSchaller:
Don't know what you mean.

JBorrow:
The observational_data git submodule needs to have its commit value revved when we add new obs data. If that's the case here, we'll need to do that or it might cause issues, so:

cd observational_data
git pull origin master
cd ..
git commit -m "Revved commit" observational_data

MatthieuSchaller:
Right. I think the data is there as it is in colibre but I'll do that nevertheless

PR #60, opened 2020-11-06
Some cold gas updates

MatthieuSchaller:
That infamous factor of 2.

### DIFF
--- a/eagle-xl/auto_plotter/cold_gas_relations.yml
+++ b/eagle-xl/auto_plotter/cold_gas_relations.yml
@@ -25,7 +25,7 @@ stellar_mass_neutral_fraction_30:
   metadata:
     title: Stellar Mass-Neutral Gas Fraction (30 kpc Stellar Mass)
     caption: All galaxies are included in the median line.
-    section: Gas Relations
+    section: Cold Gas Relations
     show_on_webpage: false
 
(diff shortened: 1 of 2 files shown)


MatthieuSchaller:
Also:

 - Add M_H2 / (M_HI + M_H2) vs. M_*
 - Use a single section for all the cold gas relation plots

JBorrow:
Do all of these make use of the adaptive binning? Not sure from the plots you've shown that they do.

MatthieuSchaller:
As far as I can tell, the scripts demand an adaptive binning. Could the odd-looking plots be due to zeros being problematic?

JBorrow:
Ah yes if the average in that bin is zero then I suppose what we see (missing bins) is expected behaviour

MatthieuSchaller:
Is there a way to fix that or should I just not plot medians?

JBorrow:
Well you could create a mask that only includes certain galaxies (e.g. those with > some_value H_2 mass) or we can just live with it.

MatthieuSchaller:
But we want zeros to be shown.

JBorrow:
In that case it is fine as-is, I suppose the median of those lines at that position _should_ be zero

MatthieuSchaller:
Then that's ready to go I think.

PR #64, opened 2020-11-16
Simplify the Z-evolution scripts and add the EAGLE-Ref data to them

MatthieuSchaller:


MatthieuSchaller:
Now also for the COLIBRE pipeline. \o/

PR #66, opened 2020-11-18
Add the critical-density line from DV&S2012 to 2 plots

EvgeniiChaikin:
**1. Why this PR** 
There was a request to add the critical-density line from DV&S2012 

**2. The definition of the critical density** 
[Dalla Vecchia, Claudio; Schaye, Joop (2012)](https://ui.adsabs.harvard.edu/abs/2012MNRAS.426..140D/abstract), eq. 18

**3. Example of the new plots** 
Two plots have been modified

- Stellar birth-density distirbution
- SNII gas-density distribution

http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/With_AGN/L006N188/With_BH_mergers/z0.0_EOS_gamma2_T8e3_n1e2_T20/

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -52,7 +52,7 @@ scripts:
     section: Metal Mass Fractions
     title: Metal Mass Fraction Distribution
   - filename: scripts/birth_density_distribution.py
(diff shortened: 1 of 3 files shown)


EvgeniiChaikin:
Note that if we compare three runs, there will be **three** lines indicating the critical density, not **one**. That is because in the most general case, the runs might have different mass resolution which would result in different critical densities.

PR #68, opened 2020-11-23
Make all HI, HII, H2 mass fraction plots compatible with CHIMES

EvgeniiChaikin:
**Why this PR**
Currently, not all (cold) gas mass-fraction plots are compatible with `CHIMES`. That's because `Table` and `CHIMES` cooling have different indices for the same species arrays.

**Note that**
We discussed it some time ago, and there has already been a fix like this one but for different plots.

### DIFF
--- a/colibre/registration.py
+++ b/colibre/registration.py
@@ -273,7 +273,7 @@ except AttributeError:
             self,
             f"gas_neutral_H_to_stellar_fraction_{aperture_size}_kpc",
             unyt.unyt_array(
-                catalogue.masses.m_gas,
+                ones,
                 name="$M_{{\\rm HI + H_2}} / M_*$ ({aperture_size} kpc) not found, showing $1$",
             ),
         )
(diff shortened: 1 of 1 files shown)


EvgeniiChaikin:
@jtrayford , can you please have a look at this PR?

wmfw23:
@EvgeniiChaikin it looks good to me, I've PR'd to your branch to make sure the `except` case for gas fractions yields dimensionless arrays as expected, but it should not enter this case now for either `Hybrid` or `Table` cooling

EvgeniiChaikin:
Indeed! The fractions should be dimensionless in the **except case** too. Thanks, @wmfw23 I merged your fix.

MatthieuSchaller:
@EvgeniiChaikin you mentioned potential worries in the XL pipeline. I think the correct selection of the fields is already done in the eagle-xl/registation.py script.

EvgeniiChaikin:
@MatthieuSchaller I still don't see how [THIS](https://github.com/SWIFTSIM/pipeline-configs/blob/4c759428abebbc38790b9e266650932d8d880108/eagle-xl/registration.py#L154-L174), which is EAGLEXL code, can get the H2 fractions from CHIMES

MatthieuSchaller:
Indeed! I only checked https://github.com/SWIFTSIM/pipeline-configs/blob/4c759428abebbc38790b9e266650932d8d880108/eagle-xl/registration.py#L301 

Thanks for being more careful than me. I'll fix it.

MatthieuSchaller:
Fixed in #76

JBorrow:
Just to be clear this is not required anymore?

MatthieuSchaller:
No, my changes were only on the XL side.

Issue #67, opened 2020-11-23
Figure Change Requests


JBorrow:
Cold Gas Relations
-------------------

Section to be handled by @jtrayford 

- [ ] Can we start the y-axis at 1e-4?

- [ ] In the captions of all plots, "fraction" is ambiguous, in the literature it can e.g. be either M_H2/Mstar or M_H2/(Mstar+M_H2) and there are more possibilities. Please detail in every caption what is shown.

- [ ] If the bottom row shows fractions of the kind X/tot instead of X/Y, then we cannot have values greater than unity. In that case please only extend the y-axis range to a value of 2 or so.

- [ ] Suggested order and type of plots for cold gas relations (every quantity is a mass):

first row: (y-axis range:
(HI+H2)/(HI+H2+star) versus star
HI/(HI+H2) vs star
H2/(HI+H2) vs star
starforming/(starforming+star) vs star

second row:
(HI+H2)/starforming vs star
H2/starforming vs star
HI/starforming vs star

third row:
(HI+H2)/star vs sSFR
HI/star vs sSFR
H2/star vs sSFR
starforming/star vs sSFR

fourth row:
(HI+H2)/(HI+H2+star) vs Sigma_star (Sigma = surface density)
HI/(HI+H2) vs Sigma_star
H2/(HI+H2) vs Sigma_star
starforming/(starforming+star) vs Sigma_star

and any plots that compare with data and are not included above.
Other plots not mentioned can be omitted

- [ ] Note, we have to clarify whether the observed gas fractions with are corrected for helium! We need to do it the same way for the simulations and specify what is done in the captions.

Density-temperature plots
---------------------------

Section to be handled by @JBorrow 

- [ ] Add a plot weighting by log dust/metals mass ratio. Only plot particle with non-zero metallicity.

Dust mass function
-------------------

Section to be handled by @JBorrow 

- [ ] Caption says it is the HI mass function.

Gas mass function
------------------

Section to be handled by @JBorrow 

- [ ] X-axis titles should specify aperture

Histograms
------------

Section to be handled by @JBorrow 

- [ ] Cumulative histograms do not need to be binned. N/A

Hydrogen phase density-temperature
--------------------------------------

Section to be handled by @JBorrow 

- [ ] Captions should specify how fraction is defined.

Metallicity
-----------

Section to be handled by @JBorrow 

- [ ] Is gas metallicity (O/H)? y-axis title suggests it is, but caption does not. It should be (O/H)

[...]

PR #69, opened 2020-11-23
Update plots/captions in colibre webpage

EvgeniiChaikin:
**1. Why this PR**
request

**2. Changes**

See [this issue](https://github.com/SWIFTSIM/pipeline-configs/issues/67)

- stellar birth density plots
- stellar mass function halo mass function plots
- star formation rate plots


EvgeniiChaikin:
**One more change:**

In all stellar mass vs halo mass plots (both EAGLE-XL and COLIBRE configs), I extended the X-axis to **lower** halo masses so that **all** objects can be seen in the plots.


**Example:**

stellar_mass_halo_mass_M200_all_100

Without this change, we would only see the objects with halo masses greater than `1e8 Msun`, which would be incomplete information.

JBorrow:
Apologies, you're going to need to fix the conflicts. Do you have a full page I can look at with these changes too?

EvgeniiChaikin:
@JBorrow 

Just a short summary of the current version of this PR

- I merged in the latest master
- the detached version of the comparison-data repository is updated, which now includes the latest changes to the EAGLE2015 and Berhoozi2019 data
- As discussed some time ago, there is no need to assign a certain SFR to the galaxies below the passive-fraction threshold. (that is why there is a small change in `registration.py`) 
- The changes in `autoplotter/stellar_mass_halo_mass.yml look horrible`. But actually, I just rearranged certain blocks of code so that the figures appear on the web-page in the order Joop was asking for. I also extended the lower limit of the halo mass in all plots with the name signatures `stellar_mass_halo_mass_*_all_30/100` so that these plots show all objects that a run may possibly have
- In `config.yml`, I added the word **birth** in the caption of stellar-birth-density figures so that it is more clear that we are talking about the birth densities and not about something else
- The new web-page will look like [THIS](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/With_AGN/L006N188/With_BH_mergers/z5.0_fkin05_EOSn1e2_vkick25_dtyoung01/)

JBorrow:
Thanks!

PR #74, opened 2020-12-03
Strip _z???p??? suffixes in multi-z obs data in SMHM plots

EvgeniiChaikin:
In this PR, I change the way multi-z observational data files are loaded for SMHM plots.

More precisely, we no longer need the suffixes `_z???p???` in Schaye15 and Berhoozi19 data files' names because multi-z data is now contained in a single `.hdf5` file.

This PR depends on [this PR](https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/54). I.e. the latter needs to be merged first.


EvgeniiChaikin:
@JBorrow I probably forgot to do one thing. To be able to use the observational data that has been updated in the last few days, we need to connect this repository to the newer version of the comparison-data repository.

EvgeniiChaikin:
I will update it in #69

PR #77, opened 2020-12-12
Add gas-part smoothing-length histogram

EvgeniiChaikin:
Add gas-particle smoothing-length histogram to the colibre pipeline

**Example**

Screenshot from 2020-12-12 17-22-31

### DIFF
--- a/eagle-xl/scripts/gas_smoothing_lengths.py
+++ b/eagle-xl/scripts/gas_smoothing_lengths.py
@@ -1,0 +1,125 @@
+"""
+Makes a histogram of the gas smoothing lengths. Uses the swiftsimio library.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from swiftsimio import load
+from unyt import unyt_quantity, dimensionless, Mpc
+
+h_bounds = [1e-2, 5e2]  # comoving kpc
+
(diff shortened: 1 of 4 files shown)


MatthieuSchaller:
I have copied over the script to the XL pipeline as well. All looks good to me. Thanks!

eagle-xl/scripts/gas_smoothing_lengths.py
```diff
@@ -0,0 +1,125 @@
...
+Makes a histogram of the gas smoothing lengths. Uses the swiftsimio library.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from swiftsimio import load
+from unyt import unyt_quantity, dimensionless, Mpc
+
+h_bounds = [1e-2, 5e2]  # comoving kpc
```
JBorrow, line 11: @MatthieuSchaller are we sure we want the same bounds for XL?

MatthieuSchaller: We use 500kpc as our max h so technically no need to go bigger. I have nevertheless increased the range a bit so the the max is seen.

MatthieuSchaller:
@JBorrow any final words?

PR #79, opened 2020-12-14
Apply the He correction also to the adaptive H_2 MF. 

MatthieuSchaller:
Mention that the correction is applied in the caption


MatthieuSchaller:
@wmfw23 let me know if you disagree with this.

wmfw23:
@MatthieuSchaller Great, this is good as far as I'm concerned, to clarify the Helium correction should be applied to all plots comparing to the XCOLDGASS survey, which applies this universally.

MatthieuSchaller:
Thanks for confirming.

PR #83, opened 2021-01-12
New dust plots

wmfw23:
Collection of new dust related plots:

- Dust-to-metal shaded phase diagram
- Hydrogen species fractions as a function of density, co-plotted with dust-to-metal ratio as a function of density
- Dust scaling relation plots fro data comparison collected from DeLooze et al (2020) data
- Dust to gas ratio as a function of oxygen abundance, to compare with Rémi-Ruyer et al (2014) data

NB. registration of some new quantities, change to VR file to add reading of mass-weighted Oxygen abundances. species fraction plot uses interpolated Ploeckinger et al (2020) tables for dust comparison. Added config arg points to an instance of this, but if unavailable will not plot. Also want to investigate selection function for DeLooze plots

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -10,7 +10,7 @@ auto_plotter_registration: registration.py
 
 # Location of the 'observational data' repository and its compiled
 # contents.
-observational_data_directory: ../../observational_data
+observational_data_directory: ../observational_data
 
 # Style sheet to be used throughout with plotting
 matplotlib_stylesheet: mnras.mplstyle
@@ -199,13 +199,6 @@ scripts:
     output_file: subgrid_density_temperature_HII.png
     section: Hydrogen Phase Density-Temperature
     title: HII
-  - filename: scripts/density_temperature_dust2metal.py
-    caption: Density-temperature diagram shaded by the total fraction of metals in the dust phase.
-    output_file: density_temperature_dust2metal.png
-    section: Density-Temperature
-    title: Dust-to-metal Ratio
-    additional_arguments:
-      quantity_type: hydro
   - filename: scripts/density_species_interp.py
     caption: Co-plot of species fractions (left y-axis) and the dust-to-metal ratio (right y-axis) as a function opf gas density. If available, solid green line shows explicitly modelled dust-to-metal ratio, while dashed green is interpolated from the Ploeckinger+20 tables.
     output_file: density_vs_species_fraction.png
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -10,7 +10,7 @@ auto_plotter_registration: registration.py
 
 # Location of the 'observational data' repository and its compiled
 # contents.
-observational_data_directory: ../observational_data
+observational_data_directory: ../../observational_data
 
 # Style sheet to be used throughout with plotting
 matplotlib_stylesheet: mnras.mplstyle
@@ -41,8 +41,6 @@ scripts:
     output_file: density_temperature_dust_to_metals.png
     section: Density-Temperature
     title: Density-Temperature (Dust / Metals)
-  - filename: scripts/density_internal_energy.py
-    caption: Density-Internal Energy diagram. If present, dashed line represents the entropy floor (equation of state).
   - filename: scripts/density_temperature_dust2metal.py
     caption: Density-temperature diagram shaded by the total fraction of metals in the dust phase.
     output_file: density_temperature_dust2metal.png
@@ -50,6 +48,8 @@ scripts:
     title: Dust-to-metal Ratio
     additional_arguments:
       quantity_type: hydro
+  - filename: scripts/density_internal_energy.py
+    caption: Density-Internal Energy diagram. If present, dashed line represents the entropy floor (equation of state).
     output_file: density_internal_energy.png
     section: Density-Temperature
     title: Density-Internal Energy
@@ -199,13 +199,6 @@ scripts:
     output_file: subgrid_density_temperature_HII.png
     section: Hydrogen Phase Density-Temperature
     title: HII
-  - filename: scripts_dev/dummy.py
-    caption: Density-temperature diagram shaded by the total fraction of metals in the dust phase.
-    output_file: density_temperature_dust2metal.png
-    section: Density-Temperature
-    title: Dust-to-metal Ratio
-    additional_arguments:
-      quantity_type: hydro
   - filename: scripts/density_temperature_dust2metal.py
     caption: Density-temperature diagram shaded by the total fraction of metals in the dust phase.
     output_file: density_temperature_dust2metal.png
@@ -221,27 +214,3 @@ scripts:
     additional_arguments:
       cooling_tables: /cosma7/data/dp004/wmfw23/colibre_dust/coolingtables/UV_dust1_CR1_G1_shield1.hdf5
       quantity_type: hydro
-  - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by H2 mass fraction.
-    output_file: subgrid_density_temperature_H2.png
-    section: Hydrogen Phases
-    title: H2 subgrid phase diagram
-    additional_arguments:
-      hydrogen_species: H2
-      quantity_type: subgrid
-  - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by HI mass fraction.
-    output_file: subgrid_density_temperature_HI.png
-    Section: Hydrogen Phases
-    title: HI subgrid phase diagram
-    additional_arguments:
-      hydrogen_species: HI
-      quantity_type: subgrid
-  - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by HII mass fraction.
-    output_file: subgrid_density_temperature_HII.png
-    section: Hydrogen Phases
-    title: HII subgrid phase diagram
-    additional_arguments:
-      hydrogen_species: HII
-      quantity_type: subgrid
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -5,7 +5,7 @@
 # relation figures. Also required is the registration file
 # in the case where you have non-catalogue properties
 # used in the autoplotter.
-auto_plotter_directory: auto_plotter
+auto_plotter_directory: auto_plotter_dev
 auto_plotter_registration: registration.py
 
 # Location of the 'observational data' repository and its compiled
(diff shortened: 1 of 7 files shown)


JBorrow:
Hi, looks like this was created against an older version of the repo. Can you fix the merge conflicts before I review?

colibre/config.yml
```diff
@@ -10,7 +10,7 @@ auto_plotter_registration: registration.py
 
 # Location of the 'observational data' repository and its compiled
 # contents.
-observational_data_directory: ../observational_data
+observational_data_directory: ../../observational_data
```
JBorrow, line 13: Why this change? Was this broken before?

colibre/config.yml
```diff
@@ -41,6 +41,13 @@ scripts:
     output_file: density_temperature_dust_to_metals.png
     section: Density-Temperature
     title: Density-Temperature (Dust / Metals)
+  - filename: scripts/density_temperature_dust2metal.py
+    caption: Density-temperature diagram shaded by the total fraction of metals in the dust phase.
+    output_file: density_temperature_dust2metal.png
+    section: Density-Temperature
+    title: Dust-to-metal Ratio
+    additional_arguments:
+      quantity_type: hydro
```
JBorrow, line 50: This appears to be repeated below

colibre/config.yml
```diff
@@ -156,27 +163,42 @@ scripts:
...
+    caption: Co-plot of species fractions (left y-axis) and the dust-to-metal ratio (right y-axis) as a function opf gas density. If available, solid green line shows explicitly modelled dust-to-metal ratio, while dashed green is interpolated from the Ploeckinger+20 tables.
+    output_file: density_vs_species_fraction.png
+    section: Hydrogen Phases
+    title: Hydrogen Phase Fractions
+    additional_arguments:
+      cooling_tables: /cosma7/data/dp004/wmfw23/colibre_dust/coolingtables/UV_dust1_CR1_G1_shield1.hdf5
+      quantity_type: hydro
   - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by H$_2$ mass fraction. The fraction is computed as the H$_2$ mass contained in each cell over the mass of gas in that cell.
+    caption: Density-temperature diagram shaded by H2 mass fraction.
```
JBorrow, line 182: Can we keep the latex-ing in these, please? So H$_2$. I was also requested to add that additional comment, so if it's incorrect can we replace it with a correct alternative?

colibre/config.yml
```diff
@@ -156,27 +163,42 @@ scripts:
...
-    section: Hydrogen Phase Density-Temperature
-    title: H$_2$
+    section: Hydrogen Phases
+    title: H$_2$ subgrid phase diagram
     additional_arguments:
       hydrogen_species: H2
       quantity_type: subgrid
   - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by HI mass fraction. The fraction is computed as the HI mass contained in each cell over the mass of gas in that cell.
+    caption: Density-temperature diagram shaded by HI mass fraction.
```
JBorrow, line 190: Similar comment above about how the binning is done.

colibre/config.yml
```diff
@@ -156,27 +163,42 @@ scripts:
...
-    section: Hydrogen Phase Density-Temperature
-    title: HI
+    section: Hydrogen Phases
+    title: HI subgrid phase diagram
     additional_arguments:
       hydrogen_species: HI
       quantity_type: subgrid
   - filename: scripts/density_temperature_species.py
-    caption: Density-temperature diagram shaded by HII mass fraction. The fraction is computed as the HII mass contained in each cell over the mass of gas in that cell.
+    caption: Density-temperature diagram shaded by HII mass fraction.
```
JBorrow, line 198: Same again here.

colibre/registration.py
```diff
@@ -111,6 +111,8 @@
     except AttributeError:
         pass
 
+metal_frac = catalogue.apertures.zmet_gas_sf_100_kpc
+nonmetal_frac = 1.-metal_frac
```
JBorrow, line 115: File needs formatting with black

colibre/registration.py
```diff
@@ -143,9 +145,46 @@
 total_dust_mass = total_dust_fraction * catalogue.masses.m_gas
 name = f"$M_{{\\rm dust}}${dust_frac_error}"
 total_dust_mass.name = name
-nonmetal_frac = 1. - catalogue.apertures.zmet_gas_sf_100_kpc
+total_dust_fraction.name = f"$\\mathcal{{DTG}}${dust_frac_error}"
+total_dust_mass = total_dust_fraction * catalogue.masses.m_gas
```
JBorrow, line 149: Should we be using the gas mass in the 100kpc aperture here?

colibre/registration.py
```diff
@@ -143,9 +145,46 @@
...
+# Get depletion properties
+try:
+    gas_mass = catalogue.masses.m_gas
+    H_frac = getattr(catalogue.element_mass_fractions, "element_0")
+    O_frac = getattr(catalogue.element_mass_fractions, "element_4")
+    
+    o_abundance = unyt.unyt_array(
+        12+np.log10(O_frac/(16*H_frac)), "dimensionless")
+    o_abundance.name = "$12+\\log_{10}({{\\rm O/H}})$" 
+    setattr(self, "gas_o_abundance", o_abundance)
```
JBorrow, line 172: Can we include aperture information here?

colibre/registration.py
```diff
@@ -143,9 +145,46 @@
...
+    
+except AttributeError:
+    # We did not produce these quantities.
+    setattr(
+        self,
+        "gas_o_abundance",
+        unyt.unyt_array(
+            catalogue.masses.m_gas**0,
```
JBorrow, line 180: not sure that this has the correct units for the 'fallback' array.

colibre/registration.py
```diff
@@ -175,6 +215,24 @@
...
+    dust_to_hi_ratio.name = "M_{{\\rm dust}}/M_{{\\rm HI}}"    
+    setattr(self, "gas_dust_to_hi_ratio", dust_to_hi_ratio)
+    
+except AttributeError:
+    # We did not produce these quantities.
+    setattr(
+        self,
+        "gas_dust_to_hi_ratio",
+        unyt.unyt_array(
+            catalogue.masses.m_gas**0,
```
JBorrow, line 230: Same comments here as above, would be nice to have the units explicit.

colibre/registration.py
```diff
@@ -175,6 +215,24 @@
         ),
     )
 
+# Get HI to dust ratio
+try:
+    dust_to_hi_ratio = total_dust_mass / HI_mass
+    dust_to_hi_ratio.name = "M_{{\\rm dust}}/M_{{\\rm HI}}"    
+    setattr(self, "gas_dust_to_hi_ratio", dust_to_hi_ratio)
```
JBorrow, line 222: Aperture information? If not in an aperture, but uses FOF halo, would be nice to have a comment here explicitly stating that.

colibre/scripts/density_species_interp.py
```diff
@@ -0,0 +1,339 @@
+"""
+Makes a rho-T plot. Uses the swiftsimio library.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import glob
+
+import swiftsimio as sw
+print(sw.__file__)
```
JBorrow, line 10: Please remove this print statement

colibre/scripts/density_species_interp.py
```diff
@@ -0,0 +1,339 @@
+"""
+Makes a rho-T plot. Uses the swiftsimio library.
```
JBorrow, line 2: Please add a longer decsription of the specific function of this rho-T instance.

colibre/scripts/density_species_interp.py
```diff
@@ -0,0 +1,339 @@
...
+    "Carbon",
+    "Nitrogen",
+    "Oxygen",
+    "Neon",
+    "Magnesium",
+    "Silicon",
+    "Sulphur",
+    "Calcium",
+    "Iron",
+    "Oatoms",]
```
JBorrow, line 38: Needs formatting with black

JBorrow, line 38: Do we not have this information from the namedcolumns instance?

colibre/scripts/density_species_interp.py
```diff
@@ -0,0 +1,339 @@
...
+    ionfrac = X * masses * data.gas.species_fractions.HII
+
+    # Dust Mass Fractions
+    dfracs = np.zeros(data.gas.masses.shape)
+    for d in dir(data.gas.dust_mass_fractions):
+        if hasattr(getattr(data.gas.dust_mass_fractions, d), "units"):
+            dfracs += getattr(data.gas.dust_mass_fractions, d)
+            
+    if glob.glob(tables):
+        table_file = h5.File(tables, "r")
```
JBorrow, line 69: Please use `with h5.File(tables, "r")` rather than explitcitly opening and closing the file. That ensures if it crashes, or an exception is raised, that the file is closed 'nicely'

colibre/config.yml
wmfw23: my mistake, I use an external data directory that i can develop separately, but need to change this back on commit

colibre/config.yml
wmfw23: removed

colibre/config.yml
wmfw23: added these back, removed accidentally

colibre/config.yml
wmfw23: as above

colibre/config.yml
wmfw23: as above

colibre/registration.py
wmfw23: done

colibre/registration.py
wmfw23: use gas mass now

colibre/registration.py
wmfw23: This is a subhalo property, added comment as below

colibre/registration.py
wmfw23: the `**0` makes it dimensionless

colibre/scripts/density_species_interp.py
wmfw23: Unfortunately, there's an extra column in the cooling tables (`Oatoms`) I thought it was more explicit to write these out rather than use named columns and add one.

colibre/registration.py
wmfw23: it is the correct units, but put an explicit conversion in each one

colibre/registration.py
wmfw23: added comment as above

colibre/scripts/density_species_interp.py
wmfw23: done

colibre/scripts/density_species_interp.py
wmfw23: changed

colibre/scripts/density_species_interp.py
wmfw23: done

colibre/scripts/density_species_interp.py
wmfw23: done

wmfw23:
overlooked this for a while, pushed new update

colibre/scripts/density_species_interp.py
JBorrow: I would prefer it if you used the named columns and added on "Oatoms", if at all possible.

wmfw23: you also need to add Sulphur & Magnesium and remove Europium to get the same columns in an order dependent way - think it's better to write out

PR #90, opened 2021-02-08
Added the EAGLE-12 SFR data ot COLIBRE pipeline

MatthieuSchaller:
@EvgeniiChaikin to help some comparisons. Anything else we want from EAGLE?

### DIFF
--- a/colibre/scripts/load_sfh_data.py
+++ b/colibre/scripts/load_sfh_data.py
@@ -289,12 +289,20 @@ def read_obs_data(path="observational_data"):
         )
     )
 
-    # Load the EAGLE NOAGN data
+    # Load the EAGLE data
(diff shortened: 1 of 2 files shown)


JBorrow (commented):
Looks good to me, let's wait on @EvgeniiChaikin

EvgeniiChaikin:
That all looks good!

Is there an SFH plot with this new curve?

MatthieuSchaller:
star_formation_history

EvgeniiChaikin:
Thanks! Wow, the EAGLE 12 is very similar to the EAGLE 25, and sometimes even above the 25. I didn't expect that.

MatthieuSchaller:
The important aspect is that it's quite noisy as the box is rather small and that whole history is only a handful of objects.

PR #96, opened 2021-02-16
Add SNII feedback density plots to colibre config

EvgeniiChaikin:
- [EXAMPLE](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/HAWK/L006N188/z1.0_01_REFERENCE/) (three new plots in the **Feedback Densities** section)

- I also extended the upper limits in the stellar birth density plots from `5e5` to `5e6` because in the lastest runs we often had data above the Y-axis range.

### DIFF
--- a/colibre/auto_plotter/feedback_densities.yml
+++ b/colibre/auto_plotter/feedback_densities.yml
@@ -1,0 +1,86 @@
+snii_thermal_feedback_density_halo_mass_max_all:
+  type: "scatter"
+  legend_loc: "upper left"
+  x:
+    quantity: "masses.mass_200crit"
+    units: Solar_Mass
+    start: 1e8
+    end: 1e13
+  y:
+    quantity: "snii_thermal_feedback_densities.max"
+    units: "mh/cm**3"
+    start: 1e-4
+    end: 5e6
+  median:
+    plot: true
+    log: true
+    adaptive: true
+    number_of_bins: 25
+    start:
+      value: 1e8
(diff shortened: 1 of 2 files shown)


EvgeniiChaikin:
[Comparison example](http://swift.dur.ac.uk/COLIBREPlots/chaikin/comparisons/HAWK/L006N188/z1.0_VARY_FKIN_AT_FIXED_THERMAL_ESNII_%5b01,19%5d/)

JBorrow:
Number of bins here doesn't really make much sense, are you really happy with 25 being the default? Normally we'd bin by every 0.2 dex, but maybe we should do something different here. Remember that this is the minimum bin width when doing adaptive stuff, too.

EvgeniiChaikin:
Yeah, indeed! It looks like the bins might too wide. I think the easiest solution would be to make it `40` instead of `25`.

This plot is 8 dex in range so our base resolution will then be `8/40 = 0.2 dex`. Do you agree @JBorrow ?

EvgeniiChaikin:
I did a few tests;  (at least) 0.2 dex per bin looks good.

JBorrow:
Yeah that sounds good to me, can you update the branch?

JBorrow:
Ah I see you have for one of the plots, but what about the other two? They have very large bins

EvgeniiChaikin:
Done; now all the three new plots have 0.2-dex bins.

PR #99, opened 2021-03-02
Add Behroozi2019 Galaxy stellar mass function

EvgeniiChaikin:
- Should be merged after https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/79
- Attach the `comparison-data submodule` to the most recent commit after the action in bullet point 1 is done
- Fix the inclusion of `Ilbert2013` data

### DIFF
--- a/colibre/auto_plotter/mass_functions.yml
+++ b/colibre/auto_plotter/mass_functions.yml
@@ -21,14 +21,7 @@ stellar_mass_function_30:
     - filename: GalaxyStellarMassFunction/Behroozi2019_all.hdf5
     - filename: GalaxyStellarMassFunction/DSouza2015.hdf5
     - filename: GalaxyStellarMassFunction/Wright2017.hdf5
-    - filename: GalaxyStellarMassFunction/Ilbert2013_z000p200.hdf5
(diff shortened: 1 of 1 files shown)


JBorrow:
There's a lot of data on these mass functions now at low-z, because all the data that gives high-z has also low-z components. Maybe we should consider adding a min/max z option for individual data sets on the config side?

EvgeniiChaikin:
> There's a lot of data on these mass functions now at low-z, because all the data that gives high-z has also low-z components. Maybe we should consider adding a min/max z option for individual data sets on the config side?

But we do have the option to specify z_min and z_max in the `convert` scripts (from raw to hdf5). Or since that is basically hard-coded in the convert scripts it is not ideal, and you think it is easier to have this functionality in the autoplotter config (as an optional argument to the comparison data), correct?

JBorrow:
Yeah I was thinking of a more 'forced' one on the autoplotter config side. That way we could only have the data we really care about at z=0.

EvgeniiChaikin:
Sure, I can add it later if you create an issue in the velociraptor repo.

Is anything to do in this PR?

PR #100, opened 2021-03-03
Correctly call the HII-colored phase plot script

EvgeniiChaikin:
Fix tiny bug in the arguments to the script

@wmfw23 can you please take a look and tell me if you agree?


wmfw23:
Great thanks, looks like it was a botched merge conflict

JBorrow:
Can you make sure that the ones in the COLIBRE repo are also fixed?

PR #102, opened 2021-03-13
Fix two small bugs with additional data points

EvgeniiChaikin:
Finally fixing this

bug1

and this

bug2

The former bug is the result of applying a mask associated with the plot's `upper` and `lower` attributes to the data points **after** the adaptive bins have been generated. In the sSFR plots, the mask `sSFR > lower = 10^{-11} 1/yr` is unnecessary. If the plot is for active-only objects, we don't have data points with `sSFR < 10^{-11} 1/yr` to begin with. The latter is because we apply `get_quantity_from_catalogue_with_mask()`, which already includes the mask with  `sSFR < 10^{-11} 1/yr`, to the raw data points before `add_median_line()` is called.

The latter bug is caused by a typo in the config file.

### DIFF
--- a/colibre/auto_plotter/star_formation_rates.yml
+++ b/colibre/auto_plotter/star_formation_rates.yml
@@ -86,9 +86,6 @@ stellar_mass_specific_sfr_all_100:
     end:
       value: 1e12
       units: solar_mass
-    lower:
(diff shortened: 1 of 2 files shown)


JBorrow:
Amazing catch!

PR #103, opened 2021-03-16
Add 2 scripts with additional information on stellar and AGN feedback

EvgeniiChaikin:
<h3> Maximal temperature vs redshift </h3>

max_temperature_redshift

<h3> Number of BH with a given total number of AGN thermal injections </h3>

num_agn_thermal_injections

<h3> Example of full page </h3>

[LINK](http://swift.dur.ac.uk/COLIBREPlots/chaikin/comparisons/HAWK/L006N188/z0.0_VARY_AGN_dT_%5b01,38%5d/)


colibre/config.yml
```diff
@@ -148,11 +148,21 @@ scripts:
     output_file: gas_smoothing_lengths.png
     section: Histograms
     title: Gas Particle Smoothing Lengths
+  - filename: scripts/number_of_agn_thermal_injections.py
+    caption: A histogram showing the cumulative number of black holes with a given total number of thermal energy injections (less than or equal to the number of particles heated) the black hole has had throughout the simulation.
```
JBorrow, line 152: Can we clarify here that the 'cumulative' is done right-to-left rather than the usual left-to-right?

colibre/config.yml
```diff
@@ -148,11 +148,21 @@ scripts:
...
+    section: Black Holes
+    title: Cumulative number of AGN thermal injections
   - filename: scripts/max_temperatures.py
     caption: Maximal temperature recorded by gas particles throughout the entire simulation.
     output_file: gas_max_temperatures.png
-    section: Histograms
+    section: Maximal Temperatures
     title: Maximal Temperature reached by gas particles
+  - filename: scripts/max_temperature_redshift.py
+    caption: A diagram displaying maximal temperatures vs redshifts at which they were reached. The pixel colour indicates the number of stellar + gas particles in the pixel.
```
JBorrow, line 162: Avoid 'a diagram' - this adds nothing. Also avoid `+` in free-text. Something like:

The maximal temperatures reached by all star and gas particles (density given by the colour map) against the redshift at which they were that temperature.

JBorrow (commented):
Only some comments on the figure captions.

colibre/config.yml
EvgeniiChaikin: Yes. Done

colibre/config.yml
EvgeniiChaikin: Agreed!

JBorrow:
Thanks

PR #104, opened 2021-03-17
Make bahamas up to date

Moyoxkit:
This adds most of the new XL plots to the bahamas config, with appropriate axis limits. The description is now also again taking out EAGLE quantities as we have changed to a new version of the code.

Also just a lot of changes that I have used for a while, but never pushed. The current bahamas config is extremely outdated.


JBorrow:
As this is all your stuff I'll merge it without review. Thanks for pushing to master, having this stuff publicly available is really useful for everyone :)

PR #105, opened 2021-03-25
Reduce the number of plots in Colibre 

EvgeniiChaikin:
We currently have too many plots (especially cold-gas-related plots)

Also,

- set the minimum number of additional points in the passive-fractions plots to **zero**
- Fix caption of the gas-mass function figure with adaptive binning


JBorrow:
Thanks!

PR #108, opened 2021-03-26
Rearrange Colibre registration file

EvgeniiChaikin:
Hi @JBorrow,

Yesterday I was going to add a new code to `registration.py` to have a new derived field. However, after looking at `registration.py` I decided to first rearrange it - otherwise, I find it too large, difficult to maintain, and bug-prone; and I don't want to make it even larger. This PR should help improve the file.

I verified that the new version works correctly by feeding the pipeline with the hawk reference run's z=0 snapshot for this branch and master branch, and then running
`diff data_master_branch.yml data_this_branch.yml`

The result is

```
3438c3438
<   x_label: $\pi R_{*, 100 {\rm kpc}}^2 / M_{*, 100 {\rm kpc}}$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
---
>   x_label: $M_{*, 100 {\rm kpc}} / \pi R_{*, 100 {\rm kpc}}^2$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
3497c3497
<   x_label: $\pi R_{*, 100 {\rm kpc}}^2 / M_{*, 100 {\rm kpc}}$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
---
>   x_label: $M_{*, 100 {\rm kpc}} / \pi R_{*, 100 {\rm kpc}}^2$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
15049c15049
<   x_label: $\pi R_{*, 100 {\rm kpc}}^2 / M_{*, 100 {\rm kpc}}$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
---
>   x_label: $M_{*, 100 {\rm kpc}} / \pi R_{*, 100 {\rm kpc}}^2$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
26601c26601
<   x_label: $\pi R_{*, 100 {\rm kpc}}^2 / M_{*, 100 {\rm kpc}}$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
---
>   x_label: $M_{*, 100 {\rm kpc}} / \pi R_{*, 100 {\rm kpc}}^2$ $\left[\frac{\rm{M}_\odot}{\rm{kpc}^{2}}\right]$
```

the difference is because I fixed a typo in the caption of one field.

Also, @MatthieuSchaller if you have any comments, let me know

### DIFF
--- a/colibre/auto_plotter/gas_mass_functions.yml
+++ b/colibre/auto_plotter/gas_mass_functions.yml
@@ -47,7 +47,7 @@ gas_H2_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_H2_mass_100_kpc"
+    quantity: "derived_quantities.gas_H2_plus_He_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -57,7 +57,7 @@ gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex (H2 masses 100 kpc).
+    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex. H$_2$ mass corrected for Helium (uses H$_2$ masses in 100 kpc apertures)
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyH2MassFunction/Fletcher2020.hdf5
@@ -67,7 +67,7 @@ adaptive_gas_H2_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_H2_mass_100_kpc"
+    quantity: "derived_quantities.gas_H2_plus_He_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -77,7 +77,7 @@ adaptive_gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width (H2 masses 100 kpc).
+    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width. H$_2$ mass corrected for Helium (uses H$_2$ masses in 100 kpc apertures)
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyH2MassFunction/Fletcher2020.hdf5
--- a/colibre/auto_plotter/gas_mass_functions.yml
+++ b/colibre/auto_plotter/gas_mass_functions.yml
@@ -3,7 +3,7 @@ gas_HI_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_HI_mass"
+    quantity: "derived_quantities.gas_HI_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -13,7 +13,7 @@ gas_HI_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: HI Mass Function
-    caption: HIMF, showing all galaxies with a fixed bin-width of 0.2 dex (uses HI masses from the whole FOF halo).
+    caption: HIMF, showing all galaxies with a fixed bin-width of 0.2 dex (HI masses 100 kpc).
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyHIMassFunction/Zwaan2003.hdf5
@@ -25,7 +25,7 @@ adaptive_gas_HI_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_HI_mass"
+    quantity: "derived_quantities.gas_HI_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -35,7 +35,7 @@ adaptive_gas_HI_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: HI Mass Function
-    caption: HIMF, showing all galaxies with an adaptive bin-width (uses HI masses from the whole FOF halo).
+    caption: HIMF, showing all galaxies with an adaptive bin-width (HI masses 100 kpc).
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyHIMassFunction/Zwaan2003.hdf5
@@ -47,7 +47,7 @@ gas_H2_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_H2_plus_He_mass"
+    quantity: "derived_quantities.gas_H2_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -57,7 +57,7 @@ gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex. H$_2$ mass corrected for Helium (uses H$_2$ masses from the whole FOF halo).
+    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex (H2 masses 100 kpc).
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyH2MassFunction/Fletcher2020.hdf5
@@ -67,7 +67,7 @@ adaptive_gas_H2_mass_function:
   legend_loc: "lower left"
   number_of_bins: 25
   x:
-    quantity: "derived_quantities.gas_H2_plus_He_mass"
+    quantity: "derived_quantities.gas_H2_mass_100_kpc"
     units: Solar_Mass
     start: 1e6
     end: 1e11
@@ -77,7 +77,7 @@ adaptive_gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width. H$_2$ mass corrected for Helium (uses H$_2$ masses from the whole FOF halo).
+    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width (H2 masses 100 kpc).
     section: Gas Mass Function
   observational_data:
     - filename: GalaxyH2MassFunction/Fletcher2020.hdf5
--- a/colibre/registration.py
+++ b/colibre/registration.py
@@ -207,9 +207,13 @@ def register_dust(self, catalogue):
                 dust_fields.append(getattr(catalogue.dust_mass_fractions, sub_path))
         total_dust_fraction = sum(dust_fields)
         dust_frac_error = ""
+
     # If the catalogue has no dust fields
     except AttributeError:
-        total_dust_fraction = np.zeros(metal_frac.size)
+        total_dust_fraction = unyt.unyt_array(
+            np.zeros(metal_frac.size),
+            units="dimensionless",
+        )
         dust_frac_error = " (no dust field)"
 
     # Label for the dust-fraction derived field
@@ -276,7 +280,7 @@ def register_hi_masses(self, catalogue):
     # Fetch gas mass
     gas_mass = catalogue.masses.m_gas
 
-    # Fetch mass fraction of H
+    # Fetch the mass fraction of Hydrogen
     H_frac = getattr(catalogue.element_mass_fractions, "element_0")
 
     # Fetch the fraction of H in HI (try CHIMES arrays first)
@@ -295,6 +299,7 @@ def register_hi_masses(self, catalogue):
     # Label of the derived field
     HI_mass.name = "$M_{\\rm HI}$"
 
+    # Register derived fields
     setattr(self, "gas_HI_mass", HI_mass)
     setattr(self, "gas_HI_plus_He_mass", HI_mass_wHe)
 
@@ -334,7 +339,7 @@ def register_dust_to_hi_ratio(self, catalogue):
         dust_to_hi_ratio = total_dust_mass / HI_mass
 
         # Register the field
-        dust_to_hi_ratio.name = "M_{{\\rm dust}}/M_{{\\rm HI}}"
+        dust_to_hi_ratio.name = "M_{\\rm dust}/M_{\\rm HI}"
         setattr(self, "gas_dust_to_hi_ratio", dust_to_hi_ratio)
 
     # Dust fields might not be present in the catalogue
@@ -348,7 +353,7 @@ def register_dust_to_hi_ratio(self, catalogue):
                 unyt.unyt_array(
                     np.ones(np.size(catalogue.masses.m_gas)), "dimensionless"
                 ),
-                name="$M_{{\\rm dust}}/M_{{\\rm HI}}$ not found, default to 1s",
+                name="$M_{\\rm dust}/M_{\\rm HI}$ not found, default to 1s",
             ),
         )
 
@@ -394,6 +399,7 @@ def register_cold_gas_mass_ratios(self, catalogue):
 
     # Fetch gas mass
     gas_mass = catalogue.masses.m_gas
+
     # Fetch hydrogen mass fraction
     H_frac = getattr(catalogue.element_mass_fractions, "element_0")
 
@@ -577,7 +583,10 @@ def register_species_fractions(self, catalogue):
     nonmetal_frac = 1.0 - catalogue.apertures.zmet_gas_100_kpc
 
     # Fetch gas mass in the apertures of 100 kpc
-    gas_mass = catalogue.apertures.mass_gas_100_kpc
+    gas_mass_100kpc = catalogue.apertures.mass_gas_100_kpc
+
+    # Fetch all gas mass
+    gas_mass = catalogue.masses.m_gas
 
     # Compute galaxy area (pi r^2)
     gal_area = (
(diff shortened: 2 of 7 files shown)


wmfw23:
this definitely seems like an improvement to me

MatthieuSchaller:
The one commit before last is necessary to get this to run at all. The same python mistake is present in master.

The last commit is necessary as this is not using the correct velociraptor names any more after the names have been made more uniform in december.

MatthieuSchaller:
Does anyone have a snapshot and VR catalogue where this works? Reading elements seems incorrect. It currently fails with:

```
/cosma/local/Python/3.6.5/lib/python3.6/site-packages/unyt/array.py:1754: RuntimeWarning: invalid value encountered in true_divide
  inp0.view(np.ndarray), inp1.view(np.ndarray), out=out_func, **kwargs
Traceback (most recent call last):
  File "/cosma/local/Python/3.6.5/bin/swift-pipeline", line 227, in <module>
    registration_file_path=registration_filename,
  File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/__init__.py", line 72, in load
    catalogue.register_derived_quantities(registration_file_path)
  File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/catalogue.py", line 412, in register_derived_quantities
    self.derived_quantities = DerivedQuantities(registration_file_path, self)
  File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/derived.py", line 96, in __init__
    self._register_quantities()
  File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/derived.py", line 110, in _register_quantities
    exec(handle.read())
  File "<string>", line 724, in <module>
  File "<string>", line 289, in register_hi_masses
AttributeError: 'Dynamic_element_mass_fractions_VelociraptorCatalog' object has no attribute 'element_0'
```

colibre/registration.py
```diff
@@ -208,6 +208,12 @@ def register_dust(self, catalogue):
         total_dust_fraction = sum(dust_fields)
         dust_frac_error = ""
 
+        if total_dust_fraction == 0:
```
EvgeniiChaikin, line 211: @MatthieuSchaller `total_dust_fraction` is a vector here, so comparing it to `int` gives an error

```
Traceback (most recent call last):
  File "/home/evgenii/anaconda3/bin/swift-pipeline", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/evgenii/pipeline/swift-pipeline", line 227, in <module>
    catalogue = load(
  File "/home/evgenii/velociraptor-python/velociraptor/__init__.py", line 72, in load
    catalogue.register_derived_quantities(registration_file_path)
  File "/home/evgenii/velociraptor-python/velociraptor/catalogue/catalogue.py", line 413, in register_derived_quantities
    self.derived_quantities = DerivedQuantities(registration_file_path, self)
  File "/home/evgenii/velociraptor-python/velociraptor/catalogue/derived.py", line 96, in __init__
    self._register_quantities()
  File "/home/evgenii/velociraptor-python/velociraptor/catalogue/derived.py", line 111, in _register_quantities
    exec(handle.read())
  File "<string>", line 724, in <module>
  File "<string>", line 213, in register_dust
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

colibre/registration.py
```diff
@@ -225,7 +225,7 @@ def register_dust(self, catalogue):
     total_dust_fraction.name = f"$\\mathcal{{DTG}}${dust_frac_error}"
 
     # Compute total dust mass
-    total_dust_mass = total_dust_fraction * catalogue.masses.m_gas
+    total_dust_mass = total_dust_fraction * catalogue.masses.mass_gas
```
EvgeniiChaikin, line 228: Isn't it better to enable backward compatibility (i.e. to first try to fetch `mass_gas` and if that is not possible then switch to `m_gas`)?

Otherwise, we will need to rerun VR for a lot of runs if we want to update plots.

colibre/registration.py
MatthieuSchaller: But it's not.  If dust is not found then you just end up with an empty list. 

This whole dust / no-dust detection is broken.

MatthieuSchaller: Note that I am working on the basis of one of the recent runs you gave me:
`/cosma7/data/dp004/jlvc76/COLIBRE/cold_gas/colibre_0023.hdf5` and 
`/cosma7/data/dp004/jlvc76/COLIBRE/cold_gas/haloes.properties.0`

colibre/registration.py
MatthieuSchaller: We really need to use the updated VR. We can't keep using the old versions; too many important updates. 
We need to rerun it at least for all the cold gas properties. 

Is it then worth making the registration more complicated by adding backward compatibility?

EvgeniiChaikin:
> Does anyone have a snapshot and VR catalogue where this works? Reading elements seems incorrect. It currently fails with:
> 
> ```
> /cosma/local/Python/3.6.5/lib/python3.6/site-packages/unyt/array.py:1754: RuntimeWarning: invalid value encountered in true_divide
>   inp0.view(np.ndarray), inp1.view(np.ndarray), out=out_func, **kwargs
> Traceback (most recent call last):
>   File "/cosma/local/Python/3.6.5/bin/swift-pipeline", line 227, in <module>
>     registration_file_path=registration_filename,
>   File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/__init__.py", line 72, in load
>     catalogue.register_derived_quantities(registration_file_path)
>   File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/catalogue.py", line 412, in register_derived_quantities
>     self.derived_quantities = DerivedQuantities(registration_file_path, self)
>   File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/derived.py", line 96, in __init__
>     self._register_quantities()
>   File "/cosma/home/jlvc76/.local/lib/python3.6/site-packages/velociraptor/catalogue/derived.py", line 110, in _register_quantities
>     exec(handle.read())
>   File "<string>", line 724, in <module>
>   File "<string>", line 289, in register_hi_masses
> AttributeError: 'Dynamic_element_mass_fractions_VelociraptorCatalog' object has no attribute 'element_0'
> ```

I tested this branch (with the version corresponding to my latest commit) on the 23rd snapshot of the L006N188 hawk reference run (can be found in `/cosma7/data/dp004/dc-chai1/HAWK/01_reference_L006N0188`). Everything ran smoothly. 

Maybe with the recent changes to the VR config, the hydrogen mass fraction is registered differently?

colibre/registration.py
EvgeniiChaikin: @MatthieuSchaller I tried to reproduce your error by running the pipeline on the snapshot in `/cosma7/data/dp004/jlvc76/COLIBRE/cold_gas/colibre_0023.hdf5`

but got a different error

```
Traceback (most recent call last):
  File "/cosma/local/Python/3.6.5/bin/swift-pipeline", line 172, in <module>
    for input, snapshot in zip(args.input, args.snapshots)
  File "/cosma/local/Python/3.6.5/bin/swift-pipeline", line 172, in <listcomp>
    for input, snapshot in zip(args.input, args.snapshots)
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/__init__.py", line 95, in load
    return SWIFTDataset(filename, mask=mask)
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 1301, in __init__
    self.get_metadata()
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 1337, in get_metadata
    self.metadata = SWIFTMetadata(self.filename, self.units)
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 173, in __init__
    self.load_particle_types()
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 395, in load_particle_types
    scale_factor=self.scale_factor,
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 634, in __init__
    self.load_metadata()
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 654, in load_metadata
    self.load_field_descriptions()
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 739, in load_field_descriptions
    self.field_descriptions = [get_desc(handle[x]) for x in self.field_paths]
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 739, in <listcomp>
    self.field_descriptions = [get_desc(handle[x]) for x in self.field_paths]
  File "/cosma/home/dp004/dc-chai1/.local/lib/python3.6/site-packages/swiftsimio/reader.py", line 731, in get_desc
    description = dataset.attrs["Description"].decode("utf-8")
AttributeError: 'str' object has no attribute 'decode'
```

Is our snapshot just a copy of one of the hawk runs' snapshots?

MatthieuSchaller: Try `/cosma7/data/dp004/jlvc76/COLIBRE/cold_gas/colibre_copy_0023.hdf5` sorry. The other one had surgery done to it. The one with copy in the same is straight taken from you. I think it's the one you were using to try to debug the restarting from snapshot.

EvgeniiChaikin: In my original halo.properties file (in `/cosma7/data/dp004/dc-chai1/HAWK/01_reference_L006N0188`), I have the fields

```
├BirthDensities_max_star  [float64: 5898] (5 attributes)
├BirthDensities_min_star  [float64: 5898] (5 attributes)
├Configuration (159 attributes)
├DensitiesAtLastSupernovaEvent_max_gas    [float64: 5898] (5 attributes)
├DustMassFractions_index_0_averagemassweighted_gas        [float64: 5898] (5 attributes)
├DustMassFractions_index_1_averagemassweighted_gas        [float64: 5898] (5 attributes)
├Efrac    [float64: 5898] (4 attributes)
├Efrac_gas        [float64: 5898] (4 attributes)
├Efrac_star       [float64: 5898] (4 attributes)
├Ekin     [float64: 5898] (4 attributes)
├ElementMassFractions_index_0_averagemassweighted_gas     [float64: 5898] (5 attributes)
├ElementMassFractions_index_4_averagemassweighted_gas     [float64: 5898] (5 attributes)
├Epot     [float64: 5898] (4 attributes)
├File_id  [int32: 1]
├ID       [uint64: 5898] (4 attributes)

```

Your catalogue file (`/cosma7/data/dp004/jlvc76/COLIBRE/cold_gas/haloes.properties.0`) has fewer fields

```
├BirthDensities_max_star  [float64: 6101] (5 attributes)
├BirthDensities_min_star  [float64: 6101] (5 attributes)
├Configuration (162 attributes)
├Efrac    [float64: 6101] (4 attributes)
├Efrac_gas        [float64: 6101] (4 attributes)
├Efrac_star       [float64: 6101] (4 attributes)
├Ekin     [float64: 6101] (4 attributes)
├Epot     [float64: 6101] (4 attributes)
├File_id  [int32: 1]
├ID       [uint64: 6101] (4 attributes)
```

I.e. the **element mass fraction** field is missing, for example. That is why the pipeline crashes with the error you reported.

EvgeniiChaikin: The disappearance of the additional gas fields might be related to the missing trailing commas in the` internal gas properties` section in the newest VR config. I add a comment on GitLab indicating the exact lines

colibre/registration.py
EvgeniiChaikin: It is a new version of VR that gives `mass_gas` in place of m_gas, correct? (i.e. not the new config file or anything else)

For the backward compatibility question, I would for @JBorrow opinion. Personally, I am okay with either way.

colibre/registration.py
EvgeniiChaikin: > But it's not. If dust is not found then you just end up with an empty list.
> 
> This whole dust / no-dust detection is broken.

Sorry, I don't understand. If there is no dust, shouldn't it err with the `AttrubuteError`, handle this error with the exception branch, and continue running?

EvgeniiChaikin: Oh, yes you are right. It is written in such a way it will never get to `getattr()` call if there is no dust. I didn't realise it. I will rewrite the code.

EvgeniiChaikin:
The current version should work with the new fields (H2 and HI masses)

The only question is  - how do we now include the Helium corrections? (in the current version of the branch, I removed all fields that had these corrections because I am not sure how to do it correctly.

Currently, we register `1 - Y - X` in apertures of a given size, and the mass-weighted `X`. So, as the zeroth-order approximation, I could take

H2_with_helium = H2 * (Y+X)_appertures / X_mass_weighed,

but I think it would be better to have X also in apertures in order to remain consistent.

MatthieuSchaller:
Thanks for all the changes. 

If we want more things, then we need to code them up as individual fields in the SWIFT outputs. 

Let's first check what we actually want before coding up 10s of fields and plots.

colibre/auto_plotter/gas_mass_functions.yml
```diff
@@ -57,7 +57,7 @@ gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex. H$_2$ mass corrected for Helium (uses H$_2$ masses from the whole FOF halo).
+    caption: H$_2$ MF, showing all galaxies with a fixed bin-width of 0.2 dex (H2 masses 100 kpc).
```
JBorrow, line 60: H$_2$

colibre/auto_plotter/gas_mass_functions.yml
```diff
@@ -77,7 +77,7 @@ adaptive_gas_H2_mass_function:
     end: 0.316 # (1e-0.5)
   metadata:
     title: H$_2$ Mass Function
-    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width. H$_2$ mass corrected for Helium (uses H$_2$ masses from the whole FOF halo).
+    caption: H$_2$ MF, showing all galaxies with an adaptive bin-width (H2 masses 100 kpc).
```
JBorrow, line 80: Same here

colibre/registration.py
```diff
@@ -24,333 +24,458 @@
...
+        setattr(self, f"gas_H2_mass_{aperture_size}_kpc", H2_mass)
 
-    setattr(self, "gas_neutral_H_mass", neutral_H_mass)
+    return
 
+
+def register_cold_gas_mass_ratios(self, catalogue):
+
+    # Define aperture size in kpc
+    aperture_sizes = [30, 100]
```
JBorrow, line 363: Can we make this a parameter to the function, seeing as we use the same apertures everywhere?

colibre/registration.py
```diff
@@ -24,333 +24,458 @@
...
+
+def register_gas_phase_metallicities(self, catalogue):
+
+    # Define aperture size in kpc
+    aperture_sizes = [30, 100]
+
+    # Metallicities relative to different units
+    solar_metal_mass_fraction = 0.0126
+    twelve_plus_log_OH_solar = 8.69
+    minimal_twelve_plus_log_OH = 7.5
```
JBorrow, line 125: Parameters like this that have fixed values should be either globals or passed as arugments to the function.

colibre/registration.py
JBorrow: Well because this requires the updated version of VR to run anyway (for the HI, H2 masses), we can saftely remove the backwards compatibility here. I don't like the lack of deprecation time, but we can't really do anything else seeing as it was so broken before.

JBorrow:
Some minor comments.

EvgeniiChaikin:
@MatthieuSchaller @JBorrow given the changes we agreed to make today, should I wait with this PR, and update it only after all the fields we need are registered by the velociraptor python library?

colibre/registration.py
```diff
@@ -24,333 +24,458 @@
...
+    return
+
+
+def register_star_metallicities(self, catalogue):
+
+    # Define aperture size in kpc
+    aperture_sizes = [30, 100]
+
+    # Solar metallicity
+    solar_metal_mass_fraction = 0.0126
```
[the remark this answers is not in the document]
EvgeniiChaikin: Done

colibre/registration.py
EvgeniiChaikin: Now these are globals

colibre/auto_plotter/gas_mass_functions.yml
EvgeniiChaikin: Fixed

MatthieuSchaller:
Thanks for the update Evgenii

JBorrow:
Yes, let's wait to merge everything once we are confident it all works together.

colibre/registration.py
```diff
@@ -24,333 +24,464 @@
...
+        )
+
+        # Label of the derived field
+        H2_mass.name = f"$M_{{\\rm H_2}}$ ({aperture_size} kpc)"
+
+        # Compute H2 mass with correction due to He
+        He_mass = getattr(catalogue.gas_H_and_He_masses, f"He_mass_{aperture_size}_kpc")
+        H_mass = getattr(catalogue.gas_H_and_He_masses, f"H_mass_{aperture_size}_kpc")
+
+        H2_mass_with_He = H2_mass * (1.0 + He_mass / H_mass)
```
EvgeniiChaikin, line 353: @wmfw23 can you please check whether this is the correct way to account for He in the masses of H2?

EvgeniiChaikin:
@MatthieuSchaller @JBorrow I am a bit concerned about the units of the metallicity fields in VR catalogue files.

Let's look at the following example below (I take a random snapshot from the hawk runs to illustrate the possible problem)


```
In [1]: from velociraptor import load

In [2]: import numpy as np

In [3]: catalogue = load("halo_0023.properties.0")

In [4]: np.max(catalogue.apertures.zmet_gas_30_kpc)
Out[4]: unyt_quantity(0.02667054, '83.33*dimensionless')

In [5]: np.max(catalogue.apertures.zmet_gas_30_kpc.to("dimensionless"))
Out[5]: unyt_quantity(2.22245577, 'dimensionless')
```

Since `zmet_gas_30_kpc` is the average metal mass fraction in the apertures, how can it be greater than 1 in dimensionless units? (here dimensionless means that it is not in the units of Z_sun \sim 0.013 \sim 1/83.33)

Is the conversion in VR config file correct?

JBorrow:
The units in the VR config file are not correct no, as they ask for the conversion to solar. Which we don't actually want.

So when we use the pipeline we use disregard_units=True (not the default for catalogue = load()) which ensures that the units are correct.

colibre/registration.py
wmfw23: Hey, yes this will do it, basically should amount to a ~ 30% increase in the H2 masses

wmfw23:
> The units in the VR config file are not correct no, as they ask for the conversion to solar. Which we don't actually want.
> 
> So when we use the pipeline we use disregard_units=True (not the default for catalogue = load()) which ensures that the units are correct.

good to know, I was wondering why these factors in the `.cfg` weren't affecting output - same goes for SFR.

EvgeniiChaikin:
> > The units in the VR config file are not correct no, as they ask for the conversion to solar. Which we don't actually want.
> > So when we use the pipeline we use disregard_units=True (not the default for catalogue = load()) which ensures that the units are correct.
> 
> good to know, I was wondering why these factors in the `.cfg` weren't affecting output - same goes for SFR.

The conversion factor for the stellar ages (in the VR config) makes a difference, though (recall that the stellar ages plot was empty for a while because the conversion of units was incorrect) @JBorrow how do we choose whether to disregard the units when reading the fields in the VR library or not?

JBorrow:
That's not correct. The reason that the stellar ages were wrong is because we had the incorrect value of `Stellar_age_input_unit_conversion_to_output_unit` (which multiplies the input value in snapshot to catalogue value) in VR terms , not `Stellar_age_to_yr` (which we read to get the units).

JBorrow:
As for your question about how we decide:

https://github.com/SWIFTSIM/velociraptor-python/blob/master/velociraptor/units.py#L102

EvgeniiChaikin:
Could someone remind me what exactly we are waiting for in this branch?

JBorrow:
You to say that it's 100% complete, and once we've merged James' changes in a rebase

EvgeniiChaikin:
I am 100 percent okay with how this branch looks except that

- There is no Fe/H plot in this branch (I guess it is OK / can be done later)
- The O/H plot is computed as the mass-weighed average, no depletion (this will have been fixed in James's following PR, right?)
- @wmfw23 , you were going to take a look at how I rewrote the registration for the H2 / HI, right? Does everyting look good there?

Otherwise the branch is good to go. I did a few tests to ensure the pipeline runs smoothly

JBorrow:
James, assign it back to me when you're happy

wmfw23:
I was happy with the H2 (https://github.com/SWIFTSIM/pipeline-configs/pull/108#discussion_r605566202), HI looks fine to me if you wanted me to check that too. Yes the idea is that the O abundance is updated by https://github.com/SWIFTSIM/pipeline-configs/pull/110 afterwards

PR #112, opened 2021-04-04
Update SF gas metallicity plot

EvgeniiChaikin:
Hi @wmfw23,

In this PR, I update the star-forming gas metallicity plot.

Instead of scaling with the solar metallicity, I now compute the O/H of the star-forming gas directly

Note that in the new plot,

- No minimum metallicity is imposed.
- Uses depleted gas metalicities

<h2>EXAMPLE</h2>

**Before**
before

**After**
after

### DIFF
--- a/colibre/auto_plotter/metallicity.yml
+++ b/colibre/auto_plotter/metallicity.yml
@@ -4,7 +4,7 @@ stellar_mass_gas_sf_metallicity_30:
   selection_mask: "derived_quantities.is_active_30_kpc"
   comment: "Active only"
   y:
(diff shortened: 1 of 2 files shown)


wmfw23:
Thanks @EvgeniiChaikin, building on this, I made https://github.com/SWIFTSIM/pipeline-configs/pull/113, which also addresses the points I made https://gitlab.cosma.dur.ac.uk/EAGLE/swift-colibre/-/merge_requests/167#note_36680,
hope its useful, lwet me know if you have any trouble

PR #113, opened 2021-04-04
WIP: Update gas metallicity plot with floors

wmfw23:
following colibre PR https://gitlab.cosma.dur.ac.uk/EAGLE/swift-colibre/-/merge_requests/168 and VR python PR https://github.com/SWIFTSIM/velociraptor-python/pull/55

@EvgeniiChaikin on top of your `update_gat_metallicity_plot` branch, building on your https://github.com/SWIFTSIM/pipeline-configs/pull/112 changes

- register sf gas O/H using mean-of-log with a floor value for particles
-  register stellar Fe/H using mean-of-log with a floor value for particles too, though not remaking @EvgeniiChaikin 's plot
-  two floor values are produced through pre-processing and picked up by VR  python, but use lower floor as default here for now (could register both)

### DIFF
--- a/colibre/auto_plotter/depletion_relations.yml
+++ b/colibre/auto_plotter/depletion_relations.yml
@@ -1,7 +1,7 @@
 oxygen_abundance_v_dust_to_gas:
   type: "scatter"
   legend_loc: "lower right"
-  selection_mask: "derived_quantities.dust_valid_abundances_100_kpc"
+  selection_mask: "derived_quantities.is_active_30_kpc"
   x:
     quantity: "derived_quantities.gas_o_abundance_100_kpc"
     units: "dimensionless"
--- a/colibre/auto_plotter/depletion_relations.yml
+++ b/colibre/auto_plotter/depletion_relations.yml
@@ -1,9 +1,9 @@
 oxygen_abundance_v_dust_to_gas:
   type: "scatter"
   legend_loc: "lower right"
-  selection_mask: "derived_quantities.is_active_30_kpc"
+  selection_mask: "derived_quantities.is_active_100_kpc"
   x:
-    quantity: "derived_quantities.gas_o_abundance_100_kpc"
+    quantity: "derived_quantities.gas_o_abundance_low_100_kpc"
     units: "dimensionless"
     start: 6.8
     end: 9.2
--- a/colibre/registration.py
+++ b/colibre/registration.py
@@ -33,6 +33,7 @@ solar_metal_mass_fraction = 0.0134
 # Solar Fe abundance (from Wiersma et al 2009a)
 solar_fe_abundance = 2.82e-5
 
+
 def register_spesific_star_formation_rates(self, catalogue, aperture_sizes):
 
     # Lowest sSFR below which the galaxy is considered passive
@@ -211,16 +212,21 @@ def register_oxygen_to_hydrogen(self, catalogue, aperture_sizes):
         O_over_H = unyt.unyt_array(np.zeros_like(gas_sf_mass), "dimensionless")
         # Avoid division by zero
         mask = gas_sf_mass > 0.0 * gas_sf_mass.units
-        O_over_H[mask] = pow(10.,log_O_over_H_times_gas_mass[mask]) / gas_sf_mass[mask]
+        O_over_H[mask] = (
+            pow(10.0, log_O_over_H_times_gas_mass[mask]) / gas_sf_mass[mask]
+        )
 
         # Convert to units used in observations
         O_abundance = unyt.unyt_array(12 + np.log10(O_over_H), "dimensionless")
-        O_abundance.name = f"SF Gas $12+\\log_{{10}}({{\\rm O/H}})$ ({aperture_size} kpc)"
+        O_abundance.name = (
+            f"SF Gas $12+\\log_{{10}}({{\\rm O/H}})$ ({aperture_size} kpc)"
+        )
 
         # Register the field
         setattr(self, f"gas_o_abundance_{aperture_size}_kpc", O_abundance)
 
     return
+
 
 def register_iron_to_hydrogen(self, catalogue, aperture_sizes):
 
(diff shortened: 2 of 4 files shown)


EvgeniiChaikin:
Thank you, @wmfw23!

You can see a new example here: http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/NEWEST/L006N188/NO_EOS/z0.0_hawk_AGN_dT_3e8_L006N188_h2_hi_version5/

Something went wrong with the dust depletion plot (there are objects with too high O/H; maybe we should add a mask?), but apart from that everything looks great!

I am now using this new version to remake the plots for all the hawk runs.

colibre/registration.py
```diff
@@ -249,30 +199,66 @@ def register_oxygen_to_hydrogen(self, catalogue, aperture_sizes):
...
+            10.0, log_Fe_over_H_times_star_mass[mask] / star_mass[mask]
+        )
+        # Convert to units used in observations
+        Fe_abundance = unyt.unyt_array(
+            np.log10(Fe_over_H / fe_solar_abundance), "dimensionless"
+        )
+        Fe_abundance.name = f"Stellar $[{{\\rm Fe/H}}]$ ({aperture_size} kpc)"
+
+        # Register the field
+        setattr(self, f"star_fe_abundance_{aperture_size}_kpc", Fe_abundance)
```
EvgeniiChaikin, line 257:
```suggestion
    # Loop over apertures
    for aperture_size in aperture_sizes:
    
        for thr, thr_label in zip(["low", "high"], ["Min = $10^{{-4}}$", "Min = $10^{{-3}}$"]):

            # Fetch Fe over H times stellar mass computed in apertures. The
            # mass ratio between Fe and H has already been accounted for.
            log_Fe_over_H_times_star_mass = getattr(
                catalogue.element_ratios_times_masses,
                f"log_Fe_over_H_times_star_mass_{thr}floor_{aperture_size}_kpc",
            )
            # Fetch stellar mass in apertures
            star_mass = getattr(catalogue.apertures, f"mass_star_{aperture_size}_kpc")
    
            # Compute stellar-mass weighted Fe over H
            Fe_over_H = unyt.unyt_array(np.zeros_like(star_mass), "dimensionless")
            # Avoid division by zero
            mask = star_mass > 0.0 * star_mass.units
            Fe_over_H[mask] = pow(
                10.0, log_Fe_over_H_times_star_mass[mask] / star_mass[mask]
            )
            # Convert to units used in observations
            Fe_abundance = unyt.unyt_array(Fe_over_H / fe_solar_abundance, "dimensionless")
            Fe_abundance.name = f"Stellar $[{{\\rm Fe/H}}]$ ({thr_label}, {aperture_size} kpc)"
    
            # Register the field
            setattr(self, f"star_fe_abundance_{thr}_{aperture_size}_kpc", Fe_abundance)

```

colibre/registration.py
```diff
@@ -249,30 +199,66 @@ def register_oxygen_to_hydrogen(self, catalogue, aperture_sizes):
...
+        # Compute stellar-mass weighted Fe over H
+        Fe_over_H = unyt.unyt_array(np.zeros_like(star_mass), "dimensionless")
+        # Avoid division by zero
+        mask = star_mass > 0.0 * star_mass.units
+        Fe_over_H[mask] = pow(
+            10.0, log_Fe_over_H_times_star_mass[mask] / star_mass[mask]
+        )
+        # Convert to units used in observations
+        Fe_abundance = unyt.unyt_array(
+            np.log10(Fe_over_H / fe_solar_abundance), "dimensionless"
```
EvgeniiChaikin, line 252: Unlike in the O/H plot, here we don't need to take `log10` (because in the stellar metallicity plot, we currently apply the `log10` to the Y-axis, not to the Y data points).

colibre/registration.py
EvgeniiChaikin: Perhaps in the above suggestion, it would be more clear to use `floor` instead of `thr`

colibre/registration.py
wmfw23: > Thank you, @wmfw23!
> 
> You can see a new example here: http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/NEWEST/L006N188/NO_EOS/z0.0_hawk_AGN_dT_3e8_L006N188_h2_hi_version5/
> 
> Something went wrong with the dust depletion plot (there are objects with too high O/H; maybe we should add a mask?), but apart from that everything looks great!
> 
> I am now using this new version to remake the plots for all the hawk runs.

Yes, I think I pushed a slightly later commit that puts a is_active mask on the depletion plot - I think the problem is that non-sf galaxies have [O/H] = 0 by default, but as you say we could at a mask to the array itself

wmfw23:
@EvgeniiChaikin implmented your suggestions and thr -> floor in `registration.py`

colibre/auto_plotter/depletion_relations.yml
```diff
@@ -1,7 +1,7 @@
 oxygen_abundance_v_dust_to_gas:
   type: "scatter"
   legend_loc: "lower right"
-  selection_mask: "derived_quantities.dust_valid_abundances_100_kpc"
+  selection_mask: "derived_quantities.is_active_30_kpc"
```
wmfw23, line 4: @EvgeniiChaikin this eliminates the 'high' [O/H] values

EvgeniiChaikin: Nice! 

- Do we want `"derived_quantities.is_active_30_kpc"` or `"derived_quantities.is_active_100_kpc"`?
- Since `"dust_valid_abundances_100_kpc"` is no longer used, do we want to keep it in the registration file?

EvgeniiChaikin:
@wmfw23, do you think we should register the O/H field with both the low and high floor (like we do for the Fe/H field), or what we have right now is already enough?

wmfw23:
attempted those fixes though not run it yet, signing out for today

PR #117, opened 2021-04-15
Add luminosity functions

MatthieuSchaller:
... and escape some of the registration when some properties don't exist.


MatthieuSchaller:
FWIW, here is the final result: http://icc.dur.ac.uk/~jlvc76/SWIFT/output_luminosity/#8707213153352484341

Needs the MR to be accepted in the vr-python, data repo and here.

PR #118, opened 2021-04-26
Add Eddington bias to stellar mass function for FLAMINGO

Moyoxkit:
Main addition is an extra derived quantity which give the stellar mass for the apertures used for FLAMINGO, with an additional error added. This moves all the masses a little bit, causing realistic scatter between bins. Implementations according to Behroozi (2019).

There is also some additional changes in here, removing useless field etc. If anything removed/changed seems odd I can revert it.


JBorrow:
Looks fine to me. You may wish to transition to the new style registration function (see COLIBRE and EAGLE) at some point.

JBorrow:
Also feel free to `git mv bahamas flamingo` if you want

PR #119, opened 2021-04-28
Add true maximum temperature to max temps plot

JBorrow:
Evgenii, could you please review these changes?

### DIFF
--- a/colibre/scripts/max_temperatures.py
+++ b/colibre/scripts/max_temperatures.py
@@ -50,7 +50,7 @@ def make_single_image(filenames, names, T_bounds, number_of_simulations, output_
     ax.text(
         0.05,
         0.05,
-        "\n".join(["Maximum $\\log_{10} T /$K"] + log_max_Ts),
+        "\n".join(["Maximum $\\log_{10} T_{\\rm max} /$K"] + log_max_Ts),
         ha="left",
         va="bottom",
         transform=ax.transAxes,
--- a/colibre/scripts/max_temperatures.py
+++ b/colibre/scripts/max_temperatures.py
@@ -33,16 +33,29 @@ def make_single_image(filenames, names, T_bounds, number_of_simulations, output_
     ax.set_ylabel("PDF [-]")
     ax.loglog()
 
+    log_max_Ts = []
+    
     for filename, name in zip(filenames, names):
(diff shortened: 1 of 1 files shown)


JBorrow:
gas_max_temperatures

colibre/scripts/max_temperatures.py
```diff
@@ -33,16 +33,30 @@ def make_single_image(filenames, names, T_bounds, number_of_simulations, output_
...
         h, bin_edges = np.histogram(
-            np.log10(T_max), range=np.log10(T_bounds), bins=250, density=True
+            log_T_max, range=np.log10(T_bounds), bins=250, density=True
         )
         bins = 0.5 * (bin_edges[1:] + bin_edges[:-1])
         bins = 10 ** bins
         ax.plot(bins, h, label=name)
 
-    ax.legend()
+        log_max_Ts.append(f"{name}: {log_T_max.max():3.3f}")
```
EvgeniiChaikin, line 48: Perhaps, no need to have such a high precision?

```suggestion
        log_max_Ts.append(f"{name}: {log_T_max.max():3.2f}")
```

colibre/scripts/max_temperatures.py
```diff
@@ -33,16 +33,30 @@ def make_single_image(filenames, names, T_bounds, number_of_simulations, output_
...
+
+    ax.text(
+        0.05,
+        0.05,
+        "\n".join(["Maximum $\\log_{10} T /$K"] + log_max_Ts),
```
EvgeniiChaikin, line 53: We are showing max of T_**max**

```suggestion
        "\n".join(["Maximum $\\log_{10} T_{\\rm max}$ [K]"] + log_max_Ts),
```

EvgeniiChaikin (commented):
Hi @JBorrow 

Just two minor, unimportant suggestions.

colibre/scripts/max_temperatures.py
JBorrow: Logarithmic transforms don't retain units, and can't be applied to quantities with units. That's why we use an algebraic expression for the units (i.e. `log X / Unit` instead of `log X [Unit]`). I will add the `_{\\rm max}` though, thanks!

PR #121, opened 2021-05-06
Update the cosmic H2 and HI density evolution plots to use the data collected for COLIBRE

MatthieuSchaller:


JBorrow:
Yes, if this requires data that is not in that sha.

JBorrow:
Looks like it already was!

PR #123, opened 2021-05-25
Zooms config

EvgeniiChaikin:
Hi,

Here is an update on zooms

<h2>Description</h2>

1) **Zooms now have their own config**. I made a copy of colibre pipeline and adopted it for zooms (excluding scripts that are volume-based such as star formation history, metal evolution, HI evolution, etc). I also excluded performance scripts because they don't make much sense for combined data.

2) **Concatenating data from multiple zooms is easy**. I wrote two python scripts that concatenate halo catalogues and snapshots into new files (one for haloes and one for snapshots). The scripts are documented and formatted so everybody can use them.

<h2>How to use</h2>

**Example 1. How to concatenate snapshots**

`concatenate_snapshots.py -p "halo_*/snapshot_halo_*_??37.hdf5" -i "/.zooms/" -o "output.hdf5" -n "Concatenated_data"`

where 

`-p` pattern to match when looking for input files. Can be, e.g., `*snapshot_halo_*_??37.hdf5` when one has many files; or one can simply write `-p snapshot_halo_01_0037.hdf5 snapshot_halo_02_0037.hdf5` to select files manually
`-i` path to directory with the zooms
`-o` output  file name
`-n` name in the header of the new file (will appear as the name of the run on a webpage)

**Example 2. How to concatenate halo catalogues**

 Everything is almost the same as with snapshots

`concatenate_catalogues.py -p "halo_*/halo_halo_*_??37.properties" -i "/.zooms/" -o "output.properties"`

This will find files such as `/.zooms/halo_1/halo_halo_1_0037.properties`. The output file will be save in the directory where the script is run.

The output file keeps a record of the files from which the data was taken. Namely, the new files will have a field
```

├ConcatenatedFiles
│ └2 attributes:
│   ├./zooms/halo_10/snapshot_halo_10_0037.hdf5: b'COLIBRE-MW'
│   └./zooms/halo_9/snapshot_halo_9_0037.hdf5: b'COLIBRE-MW'
```

<h2>Results</h2>

- halo 10 [LINK TO PLOTS](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/ZOOMS/halo_10/z0.0_REFERENCE/)
- halo 5 [LINK TO PLOTS](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/ZOOMS/halo_5/z0.0_REFERENCE/)
- halo 10 and 5 combined [LINK TO PLOTS](http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/ZOOMS/concatenated/z0.0_REFERENCE/)


colibre-zooms/concatenate_catalogues.py
```diff
@@ -0,0 +1,180 @@
...
+                        # Fetch fields
+                        concatenated_field = concatenated_data[key]
+                        new_field = data[key]
+
+                        # Compute new shape
+                        new_shape = (concatenated_field.shape[0] + new_field.shape[0],)
+
+                        # Append new data via resize method
+                        concatenated_field.resize(new_shape)
+                        concatenated_field[-new_field.shape[0] :] = new_field[:]
```
JBorrow, line 122: You may have some problems with this, as I am not sure you can ever remove a part of a HDF5 file (in python at least), it's just hidden. I would worry that when you resize it creates a new copy and leaves the old stuff dormant. Not something that _needs_ to be changed now, but if filesizes are larger than expected you'll need to combine them in memory and only write once.

JBorrow:
You may wish to remove the gas mass functions.

JBorrow:
Passive fraction plot also doesn't make much sense here

JBorrow:
Same thing for the SFR function

EvgeniiChaikin:
Thanks for the review @JBorrow !

- I just removed the gas mass function plots
- The volume-based SFR plots and the halo mass function plot have already been removed in [this commit](https://github.com/SWIFTSIM/pipeline-configs/pull/123/commits/48674489f6a1febe5219ac802715cfebd80550ab)
- Although I do agree that the passive fraction plots for individual zooms should not be compared to the data, I don't want to remove them because (1)  these plots are not volume based and it is still nice to know whether the largest object and its substructure are passive or not. (3) Also, if we feed concatenated data to the zoom pipeline (say, 10 zooms), then these plots might be even more useful.  

You can see how the webpage looks like for the most recent version here http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/ZOOMS/halo_9/z0.0_REFERENCE/

colibre-zooms/concatenate_catalogues.py
EvgeniiChaikin: Yes, that's definitely true! 

Although for halo catalogue files, that will never be a problem because memory-wise these files are negligible compared to snapshots. Next, for snapshots, I already tried to reduce the memory usage a little bit by processing only those fields that we really need.

Of course, if in the future we happen to deal with a much greater number of particles, this issue will have to be revised.

PR #124, opened 2021-06-04
Change normalisation in z-split distributions from 'per z-interval' to 'total'

EvgeniiChaikin:
In figures where we plot distributions per redshift intervals **(z>3, 1<z<3, and z<1)**, change normalization from **per z-interval** to **total**

This change solves the problem with the AGN density distribution plots.

<h2>Before</h2>

AGN_density_distribution_old

<h2>After</h2>

AGN_density_distribution

Note that the difference arises because for AGNs there are too few events at z>3. Hence, the normalization in this interval range (z>3) is a very small number which affects the y axis limits and disrupts the other two panels.

If we look at what happens, e.g., with **SN density distribution**, then the effect is minor

<h2>Before</h2>

SNII_density_distribution_old

<h2>After</h2>

SNII_density_distribution


EvgeniiChaikin:
No changes, just did a rebase to sync with the latest master

PR #130, opened 2021-06-22
Filter interlopers

james-trayford:
Add new `global_mask` to filter halos with > 10% of M200 in background particles, see PRs https://github.com/SWIFTSIM/pipeline/pull/25, https://github.com/SWIFTSIM/velociraptor-python/pull/68

### DIFF
--- a/colibre-zooms/config.yml
+++ b/colibre-zooms/config.yml
@@ -6,13 +6,13 @@
 # relation figures. Also required is the registration file
 # in the case where you have non-catalogue properties
 # used in the autoplotter.
-auto_plotter_directory: auto_temp
+auto_plotter_directory: auto_plotter
 auto_plotter_registration: registration.py
 auto_plotter_global_mask: derived_quantities.low_interloper_halos
 
 # Location of the 'observational data' repository and its compiled
 # contents.
-observational_data_directory: ../../observational_data
+observational_data_directory: ../observational_data
 
 # Style sheet to be used throughout with plotting
 matplotlib_stylesheet: mnras.mplstyle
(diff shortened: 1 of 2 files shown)


colibre-zooms/config.yml
```diff
@@ -0,0 +1,174 @@
...
+
+
+# Location of your other .yml files that are passed to the
+# velociraptor auto plotter (these describe your scaling)
+# relation figures. Also required is the registration file
+# in the case where you have non-catalogue properties
+# used in the autoplotter.
+auto_plotter_directory: auto_plotter
+auto_plotter_registration: registration.py
+auto_plotter_global_mask: derived_quantities.low_interloper_halos
```
james-trayford, line 11: @MatthieuSchaller global mask for autoplotter plots specified here

colibre-zooms/registration.py
```diff
@@ -0,0 +1,792 @@
...
+        # Register derived fields with specific star formation rates
+        setattr(self, f"specific_sfr_gas_{aperture_size}_kpc", ssfr)
+        setattr(self, f"is_passive_{aperture_size}_kpc", is_passive)
+        setattr(self, f"is_active_{aperture_size}_kpc", is_active)
+        setattr(self, f"sfr_halo_mass_{aperture_size}_kpc", sfr_M200)
+
+    return
+
+
+def register_global_mask(self, catalogue):
```
james-trayford, line 95: @MatthieuSchaller global mask is registered here

EvgeniiChaikin:
Hi @james-trayford , can you please rebase your branch onto the most recent master? (my zoom branch has now been merged)

colibre-zooms/config.yml
EvgeniiChaikin: Can you please add a comment here explaning what this global mask does?

EvgeniiChaikin (commented):
Hi @james-trayford 

Everything looks good and works as expected! I added one small comment.

The only thing I am concerned about is the criterion

`low_interloper_halos = m_bg < 0.1 * m200`

As far as I remember, this is what we agreed to use in the last telecon. However, when I was testing your branch it didn't work because the ratios `m_bh / m200` for the contaminated objects were much lower than `0.1`.

Below is an example of one of my tests

```
Python 3.8.8 (default, Apr 13 2021, 19:58:26) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from velociraptor import load

In [2]: import numpy as np

In [3]: data = load("halo_0037.properties")

In [4]: from unyt import unyt_quantity
# Select large haloes
In [5]: mask = data.masses.mass_200crit > unyt_quantity(7e11,"Msun")
# Show halo masses
In [6]: data.masses.mass_200crit[mask]
Out[6]: 
unyt_array([ 96.41186534, 146.34866986, 144.92554564, 144.65015043,
            145.88882091, 146.4368586 , 144.56002857, 145.31391723], '10000000000.0*Msun')
# Show stellar masses (one true object and a lot of contaminated ones)
In [7]: data.apertures.mass_star_100_kpc[mask]
Out[7]: 
unyt_array([1.38398504, 0.        , 0.        , 0.        , 0.        ,
            0.        , 0.        , 0.        ], '10000000000.0*Msun')
# Show contribution of bg particles
In [8]: data.bgpart_masses.bgpart_masses[mask]
Out[8]: 
unyt_array([0.        , 0.51283748, 0.38936586, 0.30083189, 0.18892573,
            0.15734631, 0.14274487, 0.11453455], '10000000000.0*Msun')

In [9]: data.bgpart_masses.bgpart_masses[mask] / data.masses.mass_200crit[mask]
Out[9]: 
unyt_array([0.        , 0.00350422, 0.00268666, 0.00207972, 0.001295  ,
            0.0010745 , 0.00098744, 0.00078819], '(dimensionless)')



```
This test is for the zoom ICs you made at Colibre fiducial resolution. All files are in `/cosma7/data/dp004/dc-chai1/zooms/01_reference/halo_1_fiducial_res/`

Do you think we can replace `0.1` with `1e-5` or something like that?

Also, I am a bit puzzled how can it be that the `M200` of the contaminated objects is very large and yet the contribution to `M200` from the background particles is very small. This is not what I had expected. @MatthieuSchaller what do you think?

MatthieuSchaller:
Recall that M200 is not computed from a number of particles in a given sphere. You build out a profile, assume it's monotonically growing, and find using interpolation the point where the mean enclosed density corresponds to x times the mean density.
If, for instance, that radial profile is not monotonic then things will go wrong and you can get basically anything as an answer for M200/R200 & friends.

If the criterion is not strong enough, could you base it on M_fof instead? This is much more objective and does not care about the particle distribution.
You could also add an additional criterion. On top of "10% of the mass" you could also demand "no more than XX BG particles". 

Does that help?

james-trayford:
@EvgeniiChaikin Yes, I noticed that I was getting very few contaminated halos with this criterion, and thought I was perhaps lucky with the halo/snapshot I was testing. I'm also a bit confused about how halos that aren't caught by this criterion can induce the large discrepancies seen in the plots.

@MatthieuSchaller So perhaps M200 is weird / ill-defined for some of the zoom halos where contamination is a factor? I will take a look using the M_fof and see if that makes more sense. I think that some form of mass fraction criteria certainly should catch the major outliers we see in the plots if things are making sense... but an N_bg criteria may work too

MatthieuSchaller:
Yes, I'd worry that a contaminated FOF object exhibits a weird profile which leads to a silly M200. The contaminant mass can then be only a tiny fraction of that. 

M_fof is more robust as it is strictly based on the particles.

I agree, if it works, a mass-based criterion would be better than a number-based one.

james-trayford:
rebased and updated the filter criteria to use Mfof instead of M200

EvgeniiChaikin:
Great! @james-trayford I have just run this new version on my test example. Now everything works as expected and all the contaminated haloes are gone!

As far as I know, the only thing left is to do `rebase` in https://github.com/SWIFTSIM/pipeline/pull/25 (swift-pipeline)

PR #136, opened 2021-07-13
Add stellar abundances

correac:
WiP: Adding [O/Fe] and [Mg/Fe] plots.

When trying to run the pipeline I get this error.

yaml.parser.ParserError: while parsing a block mapping
  in "../pipeline-configs/colibre/config.yml", line 243, column 5
expected <block end>, but found '<scalar>'
  in "../pipeline-configs/colibre/config.yml", line 244, column 21

which I do not know how to solve. Any suggestion?


bwvdnbro:
As @JBorrow already pointed out in https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/93#issuecomment-879671018, you should really try to avoid reading raw data in the scripts, and use a conversion script in https://github.com/SWIFTSIM/velociraptor-comparison-data instead. Most of the code in the `read_obs_data*` routines in your script can be moved to a conversion script. Within that script, you can then properly keep track of the references, instead of trying to contain all that information within the figure caption. I suspect that could help with the `ParserError` as well.

correac:
Ok! @bwvdnbro could you please make the conversion script to fix the ParserError?

bwvdnbro:
Yes, I will have a look tomorrow. I'm still getting up to speed with how the pipeline works.

correac:
Thank you so so much! Yes, no hurry.

bwvdnbro:
@correac final question: could you please clarify to me what is happening with the Koch (2008) data file? It is read twice. The first time, all lines are read and assigned to a list labelled "Carina" (is this the Kock+(05) reference in the caption?) and the second time, 2 lines are skipped and the rest is converted and added to the Milky Way list. One of them seems wrong.

correac:
Hi Bert @bwvdnbro this is indeed a mistake. Koch+08 only corresponds to Carina, it should not be added as data point for MW.

PR #137, opened 2021-07-28
Added (working but unpolished) scripts to calculate hot has properties.

jemme07:
Added (working but unpolished) scripts to calculate hot has properties for Flamingo, e.g. density, pressure and entropy profiles, etc. 

This need a lot of housekeeping! Unfortunately, I haven’t much time on my hands at the moment. @bwvdnbro, could you help with this? 

Everything should work, but it needs tidying up. 

It requires the observational data (also needs tidying up) from the fork https://github.com/jemme07/velociraptor-comparison-data.


bwvdnbro:
@jemme07 could you remove the `__pycache__` files from your repository?

bwvdnbro:
And please open a pull request for the changes in the observational data as well, that makes it easier to keep track of things.

jemme07:
Opened a pull request for the changes in the observational data as well. -> DONE

jemme07:
> @jemme07 could you remove the `__pycache__` files from your repository?

Done

flamingo/scripts/interpolate_X_Ray_redshift.py
```diff
@@ -0,0 +1,372 @@
...
+                             3.8801527e-03]
+    abundance_to_solar = 1 - mass_fraction / solar_mass_fractions
+
+    # Calculate abundance offsets using solar abundances
+    # abundance_to_solar = 1 - abundances / 10**interp.solar_metallicity
+
+    abundance_to_solar = np.c_[abundance_to_solar[:, :-1], abundance_to_solar[:, -2], abundance_to_solar[:, -2], abundance_to_solar[:, -1]] #Add columns for Calcium and Sulphur and add Iron at the end
+
+    #Find helium offsets
+    idx_he = find_idx_he(np.log10(abundances[:, 1]), interp.He_bins)
```
bwvdnbro, line 358: @jemme07 There is an inconsistency here: while `idx_n` and `idx_T` are determined using the mask `joint_mask`, `idx_he` is not. As a result, `idx_n/idx_T` and `idx_he` do not have the same length. This does not give any errors because you use the shorter length (that of `idx_n`) to determine the loop count in `get_table_interp()`, but it does mean that there is a mismatch between the elements in `idx_n/idx_T` and `idx_he`. In other words: you are probably using the wrong `idx_he` for most of the elements during the interpolation.

jemme07: The interpolation code was written by Joey Braspenning. I'll contact him to take a look. Well spotted!

flamingo/scripts/hot_gas_plots.py
```diff
@@ -0,0 +1,460 @@
...
+        df_gas['mass'] = data.gas.masses.to('Msun').value.astype(np.float64) * u.Msun.to(u.gram)
+        # df_gas['metallicity'] = data.gas.metal_mass_fractions.value / 0.0134
+        df_gas['sfr'] = data.gas.star_formation_rates.to('Msun/Gyr').value
+        df_gas['rho'] = data.gas.densities.to('g * cm ** -3').value
+        df_gas['nH'] = df_elements.hydrogen.values * df_gas.rho.values / mp 
+        # df_gas['nH'] = np.full_like(df_gas.temp.values, 1e6)
+        df_gas['lambda_c'] = np.power(10, xray.interpolate_X_Ray_pandas(np.log10(df_gas.nH.values), np.log10(df_gas.temp.values), df_elements, z, obs_data_dir, band = '0.5-2.0keV', fill_value = -400)) / df_gas.nH.values ** 2
+        df_gas['in_FOF'] = mask
+
+        df_elements = df_elements.div(df_elements.hydrogen, axis=0)
```
bwvdnbro, line 121: @jemme07 it looks like you are only setting the mass fractions of sulphur and calcium after you used the elemental mass fractions in the X-ray interpolation on line 118. Is this correct? Right now the interpolation is using the same mass fraction as for silicon for these elements. Do the tables already assume the same sulphur/silicon and calcium/silicon?

flamingo/scripts/hot_gas_plots.py
```diff
@@ -0,0 +1,460 @@
...
+        z = meta.redshift
+        ez = cosmo.efunc(z)
+
+        df_elements = pd.DataFrame()
+        for element in elements:
+            df_elements[element] = getattr(data.gas.smoothed_element_mass_fractions, element)
+
+        df_gas = pd.DataFrame(data.gas.coordinates.value, columns=['x', 'y', 'z'])
+        df_gas['temp'] = data.gas.temperatures.value
+        df_gas['mass'] = data.gas.masses.to('Msun').value.astype(np.float64) * u.Msun.to(u.gram)
```
bwvdnbro, line 112: Is there a specific reason why you are not simply using `data.gas.masses.to("gram").value`? Is this a rounding issue?

flamingo/scripts/hot_gas_plots.py
MatthieuSchaller: Yes, that is a standard assumption throughout EAGLE&co. Worth adding a comment though.

flamingo/scripts/dummy.py
```diff
@@ -0,0 +1,3 @@
+"""
+Dummy
+"""
```
JBorrow, line 3: Is there a particular reason that this exists?

flamingo/scripts/hot_gas_plots.py
jemme07: Yes, as @MatthieuSchaller mentioned, this is a standard practice.

flamingo/scripts/hot_gas_plots.py
jemme07: Yes, the problem is that the values are really small in grams, and rounding issues arise further down in the calculations and would return either inf of nan. 

I tried different hacks here, but if I remember correctly, the only way I made this work is to convert to Msun first using float64, the convert to grams. But there might be a better solution.

flamingo/scripts/dummy.py
jemme07: This dummy was aded as a hacky work around to produce several plots with the same script. That is, we read the particle data only once, make all the calculations (x-ray, entropy, etc) and produce all the plots. Then call the  dummy script in the pipeline config.yml to properly link the plots and names in the generated webpage.

bwvdnbro: The dummy script was removed in https://github.com/jemme07/pipeline-configs/pull/1, since @EvgeniiChaikin implemented a feature that allows creating multiple figures with a single script (https://github.com/SWIFTSIM/pipeline/pull/26).

flamingo/scripts/hot_gas_plots.py
```diff
@@ -0,0 +1,460 @@
...
+                df_i['volume'] = (4./3) * np.pi * (((df_i.r500 * aexp * u.Mpc.to(u.cm) * r_bins[j + 1]) ** 3.) - ((df_i.r500 * aexp * u.Mpc.to(u.cm) * r_bins[j]) ** 3.))
+                df_i['rho_tot'] = df_i.mass / df_i.volume 
+                df_i['x_i'] = df_i.ne_nH / ((df_i.ne_nH + df_i.ni_nH) * df_i.mu * mp)
+                df_i['ne'] = df_i.rho_tot * df_i.x_i
+                df_i['wtemp'] = df_i.temp * df_i.mass
+                groups = df_i.groupby(['GroupNumber', 'm500'])
+                kT = kb / ergs_keV * groups.wtemp.sum() / groups.mass.sum()
+                p = kT * groups.ne.sum()
+                s = kT / groups.ne.sum() ** (2./3)
+                p /= groups.head(1).P500.median()
```
bwvdnbro, line 347: @jemme07 I'm not sure I understand what you are doing here. Looks like you are constructing radial bins and then finding all halos that have particles within a bin. For those particles, you then compute the (total?) pressure and entropy within the bin and then you divide this by the median P500 and S500 of all halos that have particles in this particular bin. Is that right?

PR #145, opened 2021-09-24
LOS stellar velocity dispersion

bwvdnbro:
The stellar velocity dispersion output by VR is a 3D velocity dispersion. To compare this with observations, we should convert it to a LOS velocity dispersion, which can be done by multiplying with 1/sqrt(3).

The new LOS velocity dispersion shifts the simulation data in the Stellar Velocity Dispersion-Black Hole Mass relation, e.g.:
old plot:
old_stellar_veldisp_black_hole_mass_10
new plot:
new_stellar_veldisp_black_hole_mass_10

A new Stellar Mass-Stellar Velocity Dispersion plot was added:
new_stellar_mass_stellar_veldisp_10
In this plot, the new LOS velocity dispersion is used. Both the velocity dispersion and the mass are measured in the same 10 kpc aperture.

### DIFF
--- a/colibre/auto_plotter/black_holes.yml
+++ b/colibre/auto_plotter/black_holes.yml
@@ -154,7 +154,7 @@ stellar_veldisp_black_hole_mass_30:
       value: 1e3
       units: km/s
   metadata:
-    title: Stellar Velocity Dispersion-Black Hole Mass relation (30 kpc)
+    title: LOS Stellar Velocity Dispersion-Black Hole Mass relation (30 kpc)
     caption: The 3D stellar velocity dispersion is converted into a LOS velocity dispersion using a $1/\sqrt{3}$ correction factor. Note that the stellar velocity dispersion in observations is measured in a fixed 1 kpc aperture.
     section: Black Holes
     show_on_webpage: false
@@ -186,7 +186,7 @@ stellar_mass_stellar_veldisp_10:
       value: 1e12
       units: Solar_Mass
   metadata:
-    title: Stellar Mass-Stellar Velocity Dispersion relation (10 kpc)
+    title: Stellar Mass-LOS Stellar Velocity Dispersion relation (10 kpc)
     caption: The 3D stellar velocity dispersion is converted into a LOS velocity dispersion using a $1/\sqrt{3}$ correction factor.
     section: Black Holes
     show_on_webpage: true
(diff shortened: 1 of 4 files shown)


colibre/registration.py
```diff
@@ -762,6 +765,15 @@ def register_stellar_birth_densities(self, catalogue):
     return
 
 
+def register_los_star_veldisp(self, catalogue):
+    for aperture_size in [10, 30]:
+        veldisp = getattr(catalogue.apertures, f"veldisp_star_{aperture_size}_kpc")
+        los_veldisp = veldisp / np.sqrt(3.0)
```
EvgeniiChaikin, line 771: Let's add a name to this new field; otherwise, in the figures, it might be unclear what is plotted. See, e.g., x axis in your example

xaxis


```suggestion
        los_veldisp = veldisp / np.sqrt(3.0)
        los_veldisp.name = f"LOS stellar velocity dispersion ({aperture_size} kpc)"
```

colibre/auto_plotter/black_holes.yml
```diff
@@ -123,7 +123,7 @@ stellar_veldisp_black_hole_mass_10:
       units: km/s
   metadata:
     title: Stellar Velocity Dispersion-Black Hole Mass relation (10 kpc)
```
EvgeniiChaikin, line 125: I think it would be good to indicate that the plotted velocity dispersion is the **LOS** velocity dispersion.

If you agree with this suggestion, then please also update the titles of the other figures where the LOS velocity dispersion is shown.

```suggestion
    title: LOS Stellar Velocity Dispersion-Black Hole Mass relation (10 kpc)
```

EvgeniiChaikin (commented):
Hi @bwvdnbro ,

Everything looks good! I successfully ran your branch on one of my examples.

There are two small comments that I've added to the code diff. Please take a look.

After those two threads are resolved, let's also add your update to `colibre-zoom`, and then merge this PR.

bwvdnbro:
Thanks for the review! I have made the changes you requested.

I think Joop wants us to start using smaller apertures for the velocity dispersion at some point. But maybe it is better to do this separately later on, since this requires a rerun of VR.

EvgeniiChaikin:
@bwvdnbro Everything looks perfect now! Could you merge this branch to master? (I don't have enough rights to do it :) ) Also, you might need to rebase first.

Yes, I agree. Let's switch to smaller apertures later.

PR #147, opened 2021-10-07
Add the requested stellar abundances scripts with O/H along the X axis

EvgeniiChaikin:
Add three requested stellar abundances scripts to colibre pipeline with O/H along the X axis.

- O/Fe vs O/H
- Mg/Fe vs O/H
- Fraction of Fe from SNIa vs O/H

Examples of the new plots can be found here: http://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/HAWK/L012N188/z0.0_106b3_104b2_norm_0p3_new_cooling_L012N188/index.html

Note that the new code is very similar to that of the scripts with Fe/H along the X-axis. 

Also note that if we will need more scripts like this in the future, I am most likely going to reduce the number of files by making single files produce multiple plots. That is because having repeated code in multiple files is considered bad practice, so I want to avoid it wherever possible.

### DIFF
--- a/colibre/scripts/stellar_abundances_FeH_Fe_snia_fraction.py
+++ b/colibre/scripts/stellar_abundances_FeH_Fe_snia_fraction.py
@@ -20,9 +20,9 @@ def read_data(data: swiftsimio.SWIFTDataset) -> Tuple[np.ndarray, np.ndarray]:
     mFe_in_cgs = 55.845 * unyt.mp
 
     # Asplund et al. (2009)
-    Fe_H_Sun = 7.5
+    Fe_H_Sun_Asplund = 7.5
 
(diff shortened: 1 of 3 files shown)


bwvdnbro:
I agree that it would be better to have one script that produces these plots, especially since there are some hard-coded values that need to be consistent between different scripts. Should I open an issue about this, so that we do not forget?

JBorrow:
Don't we already have that feature? James used it before.

EvgeniiChaikin:
> I agree that it would be better to have one script that produces these plots, especially since there are some hard-coded values that need to be consistent between different scripts. Should I open an issue about this, so that we do not forget?

@bwvdnbro I have added this task to my task list; I am going to submit a PR next week.

PR #151, opened 2021-11-19
Added dead time plots to run performance section.

bwvdnbro:
Recent versions of SWIFT output an additional column in timesteps.txt that quantifies the dead time during every step. This pull request adds two plots to the run performance section of the pipeline that use this information.

The first plot is a dead time fraction per time-bin plot:
deadtime_timebin_hist

The second one is a dead time fraction per number of updates in the step plot:
DMO_stepsize_deadtime

In the second plot, only steps where nothing special happens (e.g. a rebuild, snapshot output...) are included, which explains the apparent discrepancy between the two plots.

The scripts detect when an old timesteps.txt file is used that does not have the extra column, and simply add an empty line to the plot with an additional message in the label warning about the missing data.

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -198,6 +198,16 @@ scripts:
     title: Wall-clock time per time-bin
     section: Run Performance
     output_file: wallclock_timebin_hist.png
+  - filename: scripts/deadtime_timebin_hist.py
+    caption: Dead time fraction in each simulation time-bin.
+    title: Deadtime fraction per time-bin
+    section: Run Performance
+    output_file: deadtime_timebin_hist.png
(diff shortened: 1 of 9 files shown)


MatthieuSchaller:
Thanks. Makes me think we should have a category for joint plots somewhere rather than duplicating the scripts for each subgrid model.

bwvdnbro:
I totally agree with that.

It would be quite easy to share the scripts (just move them to a separate folder and link to them from all models), but we probably want to be able to do this for the shared config and autoplotter files as well, which would require a change to the pipeline. Do @JBorrow or @EvgeniiChaikin have any thoughts on this?

PR #154, opened 2022-01-09
The halo mass function should only consider central haloes

MatthieuSchaller:


colibre/auto_plotter/halo_mass_functions.yml
```diff
@@ -18,3 +19,25 @@ halo_mass_function:
...
+
+halo_mass_function_adaptive:
+  type: "adaptivemassfunction"
+  select_structure_type: 10  
+  legend_loc: "lower left"
+  number_of_bins: 25
+  x:
+    quantity: "masses.mass_200crit"
+    units: Solar_Mass
+    start: 1e9
```
EvgeniiChaikin, line 31: Can you please use **the same** axis limits in `halo_mass_function_adaptive` plot and `halo_mass_function` plot?

colibre/auto_plotter/halo_mass_functions.yml
```diff
@@ -1,5 +1,6 @@
 halo_mass_function:
   type: "massfunction"
+  select_structure_type: 10
```
EvgeniiChaikin, line 3: Can you please also include `comment: "Centrals only"` for every plot that now has `select_structure_type: 10` so that we know that the plot shows only central objects

EvgeniiChaikin (commented):
Thanks, @MatthieuSchaller . I just added two small comments

colibre/auto_plotter/halo_mass_functions.yml
MatthieuSchaller: Good point.

colibre/auto_plotter/halo_mass_functions.yml
MatthieuSchaller: I can but that should be obvious for a halo mass function.

MatthieuSchaller:
What do you think of the new version?

colibre/auto_plotter/halo_mass_functions.yml
EvgeniiChaikin: You are right. However, I mentioned it because the pattern followed in the pipeline is that (almost) every plot with the mask `select_structure_type: 10` has a comment "Centrals only", so we always know exactly what is plotted.

EvgeniiChaikin:
Thanks for the changes! The new version looks good to me. I don't have much else to add.

MatthieuSchaller:
Thanks for checking.

PR #160, opened 2022-02-11
New radio data in star formation history plot

bwvdnbro:
Adds recent radio SFRD data from Novak et al. (2017), Gruppioni et al. (2020) and Enia et al. (2022) to the star formation history plot. Also fixes an issue with the observational data sub-module not being up-to-date with the latest version of the pipeline scripts.

I have only changed this for the COLIBRE pipeline for now, but I assume the same data needs to be added to the other projects. We will also need to remove some older datasets, since the plot has become unreadable. I will update the other projects once we are happy with the new COLIBRE version.

### DIFF
--- a/colibre/scripts/star_formation_history.py
+++ b/colibre/scripts/star_formation_history.py
@@ -161,7 +161,7 @@ for radio_data, color in zip(radio_data_sample, ["maroon", "goldenrod", "slategr
             zorder=-10,
         )
     )
-    observation_labels.append(radio_data.citation)
+    observation_labels.append(f"{radio_data.citation} [radio]")
 
 # Add Behroozi data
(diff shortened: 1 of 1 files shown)


EvgeniiChaikin (approved):
Hi @bwvdnbro ,

I pushed two commits to your branch.

- One to rebase to the most recent master
- And the other to add unique colours to the new data (otherwise it will have the same colours as some other data). I also slightly changed zorder and notation in the code

This branch now looks ready to me to be merged.

The SFH plot with the new changes will look as follows
star_formation_history

bwvdnbro:
Thanks! I will make some more changes to add extra `[radio]` labels to the new data. And I'm still waiting to see if some of the old data can be removed.

PR #161, opened 2022-02-15
Extend plot ranges fix small bugs

EvgeniiChaikin:
- Extend plot ranges in plots with gas density
- Fix **KeyError** bug in density distribution plots (missing attribute)
- Format with `python3.8 -m black`

### DIFF
--- a/colibre-zooms/scripts/birth_density_distribution.py
+++ b/colibre-zooms/scripts/birth_density_distribution.py
@@ -5,6 +5,7 @@ Plots the birth density distribution.
 import matplotlib.pyplot as plt
 import numpy as np
 import unyt
+import traceback
 
 from unyt import mh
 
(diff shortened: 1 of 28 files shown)


EvgeniiChaikin:
**An example of a test of the branch with the changes here:**

https://swift.dur.ac.uk/COLIBREPlots/ploeckinger/COLIBRE_highrestest/comparison_6Mpcboxes/z2.00/

colibre-zooms/scripts/birth_density_distribution.py
```diff
@@ -122,17 +123,42 @@ def critical_density_DVS2012(
...
+            N_ngb=N_ngb_target,
+            X_H=X_H,
+        )
 
-    n_crit = critical_density_DVS2012(
-        T_K=SNII_heating_temperature, M_gas=M_gas.value, N_ngb=N_ngb_target, X_H=X_H
-    )
+    # Cannot find argument(s)
+    except KeyError:
+        print(traceback.format_exc())
```
bwvdnbro, line 159: Is this still useful or is this a leftover from debugging?

colibre-zooms/scripts/last_SNII_density_distribution.py
```diff
@@ -165,17 +166,45 @@ def critical_density_DVS2012(
...
+            N_ngb=N_ngb_target,
+            X_H=X_H,
+        )
 
-    n_crit = critical_density_DVS2012(
-        T_K=SNII_heating_temperature, M_gas=M_gas.value, N_ngb=N_ngb_target, X_H=X_H
-    )
+    # Cannot find argument(s)
+    except KeyError:
+        print(traceback.format_exc())
```
bwvdnbro, line 202: Same here.

colibre/scripts/birth_density_distribution.py
```diff
@@ -122,17 +123,42 @@ def critical_density_DVS2012(
...
+            N_ngb=N_ngb_target,
+            X_H=X_H,
+        )
 
-    n_crit = critical_density_DVS2012(
-        T_K=SNII_heating_temperature, M_gas=M_gas.value, N_ngb=N_ngb_target, X_H=X_H
-    )
+    # Cannot find argument(s)
+    except KeyError:
+        print(traceback.format_exc())
```
bwvdnbro, line 159: And here.

colibre/scripts/last_SNII_density_distribution.py
```diff
@@ -165,17 +166,42 @@ def critical_density_DVS2012(
...
+            N_ngb=N_ngb_target,
+            X_H=X_H,
+        )
 
-    n_crit = critical_density_DVS2012(
-        T_K=SNII_heating_temperature, M_gas=M_gas.value, N_ngb=N_ngb_target, X_H=X_H
-    )
+    # Cannot find argument(s)
+    except KeyError:
+        print(traceback.format_exc())
```
bwvdnbro, line 202: And here.

bwvdnbro (approved):
Looks good to me. I put one comment (actually 4 times the same one) about some leftover debugging code of which I'm not sure we want to keep it.

We should probably consider creating a formatting script to avoid too many formatting changes that clutter the diff.

colibre-zooms/scripts/birth_density_distribution.py
EvgeniiChaikin: I decided to keep it purposely. I think it is useful to have so that whenever there is a key error, we will see it immediately in the output.

If you think that these lines make the code less readable or/and that we should not have any explicit traceback references and output (unless the pipeline is run in the debug mode), let me know and I will remove the lines (in this and in the other files)

EvgeniiChaikin:
> Looks good to me. I put one comment (actually 4 times the same one) about some leftover debugging code of which I'm not sure we want to keep it.
> 
> We should probably consider creating a formatting script to avoid too many formatting changes that clutter the diff.

Thanks! I agree about the formatting. At the moment it is basically arbitrary.

colibre-zooms/scripts/birth_density_distribution.py
bwvdnbro: That's fine with me then. Let's keep it.

PR #164, opened 2022-03-14
Added tick marks to top and right axis in default plotting style

bwvdnbro:
This does not affect plots that use a different style or that use custom ticks.


EvgeniiChaikin (approved):
Everything looks good, thanks!

E.g.

ticks_example

PR #166, opened 2022-03-22
Replaced 100kpc apertures with 50kpc apertures in COLIBRE pipeline.

bwvdnbro:
The old 100 kpc apertures are still there, but they are no longer shown on the web page.


EvgeniiChaikin:
Thank you @bwvdnbro !

I have one question:

With your changes, each pipeline run will now produce over 300 plots, which is too many in my opinion. Would it be better to just replace 100kpc by 50kpc? Then we won't be creating any extra plots. 

If you are okay with this, I will push this change to your branch and then we can merge it.

EvgeniiChaikin:
Hi @bwvdnbro just letting you know that I don't plan to add anything else here in case you are waiting for me.

PR #167, opened 2022-03-22
Added plot of the dead time fraction as a function of scale factor.

bwvdnbro:
The new plot looks like this:
image

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -208,6 +208,11 @@ scripts:
     title: Step size vs deadtime fraction
     section: Run Performance
     output_file: stepsize_deadtime.png
+  - filename: scripts/deadtime_evolution.py
+    caption: Evolution of the dead time fraction as a function of scale factor.
+    title: Dead time fraction vs scale factor
+    section: Run Performance
+    output_file: deadtime_evolution.png
   - filename: scripts/density_temperature_species.py
     caption: Density-temperature diagram shaded by H$_2$ mass fraction. The fraction is computed as the H$_2$ mass contained in each cell over the mass of gas in that cell.
     output_file: subgrid_density_temperature_H2.png
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -209,7 +209,7 @@ scripts:
     section: Run Performance
     output_file: stepsize_deadtime.png
   - filename: scripts/deadtime_evolution.py
-    caption: Evolution of the dead time fraction as a function of scale factor.
+    caption: Evolution of the dead time fraction as a function of scale factor. The dashed line represents the average dead time over the entire run.
     title: Dead time fraction vs scale factor
     section: Run Performance
     output_file: deadtime_evolution.png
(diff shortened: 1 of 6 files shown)


colibre/scripts/deadtime_evolution.py
```diff
@@ -0,0 +1,90 @@
...
+
+run_names = arguments.name_list
+run_directories = [f"{directory}" for directory in arguments.directory_list]
+output_path = arguments.output_directory
+
+plt.style.use(arguments.stylesheet_location)
+
+fig, ax = plt.subplots()
+
+color_index = 0
```
EvgeniiChaikin, line 25: Why not use `enumerate()` for `color_index`?

colibre/scripts/deadtime_evolution.py
```diff
@@ -0,0 +1,90 @@
+"""
+Plots dead time fraction v.s. step size
```
EvgeniiChaikin, line 2: What you probably want to write here is, e.g.,

_Plots dead time fraction v.s. cosmic scale factor_

colibre/config.yml
```diff
@@ -208,6 +208,11 @@ scripts:
     title: Step size vs deadtime fraction
     section: Run Performance
     output_file: stepsize_deadtime.png
+  - filename: scripts/deadtime_evolution.py
+    caption: Evolution of the dead time fraction as a function of scale factor.
```
EvgeniiChaikin, line 212: Maybe indicate in the caption what the dashed line stands for? (i.e. the mean dead time for the whole run)

colibre/scripts/deadtime_evolution.py
```diff
@@ -0,0 +1,90 @@
...
+            loose=True,
+            invalid_raise=True,
+            usecols=(2, 12, 14),
+            dtype=[
+                ("a", "f4"),
+                ("wallclock", "f4"),
+                ("deadtime", "f4"),
+            ],
+        )
+    except:
```
EvgeniiChaikin, line 59: Maybe it is worth being more specific and writing here which exceptions we expect to catch here? I guess? `FileNotFoundError` and `ValueError`?

EvgeniiChaikin (approved):
Thank you for the update @bwvdnbro ! Everything looks good and works as expected!

I left a few very minor suggestions that you might consider including.

colibre/scripts/deadtime_evolution.py
bwvdnbro: Because I copied this script from `stepsize_deadtime.py`, where it is also done like that. Now that I think about it, I also wrote that script, and you did not review that one. So that is probably why.

colibre/scripts/deadtime_evolution.py
bwvdnbro: It is mostly meant to catch `ValueError`s because of old timestep files that do not have the deadtime column. I realised that also means you need to set `loose=False` in the function call, so I have updated that as well.

PR #169, opened 2022-04-05
Log-normal stellar mass scatter

bwvdnbro:
This PR adds a new derived quantity to the catalogue, `mass_star_with_scatter_50_kpc`, that adds a random 0.3 dex log-normal scatter to the 50 kpc aperture stellar masses. These are then used to create alternative plots of the stellar mass function and passive fraction - stellar mass relation.

I have only tested this for a 6Mpc box, so the examples are not very illustrative.
Versions without scatter:

Versions with scatter:


bwvdnbro:
The `black` version on the CI runner seems to be broken, so the check keeps failing. Will try again later.

bwvdnbro:
I have now fixed the CI. @EvgeniiChaikin any thoughts on this PR?

PR #170, opened 2022-04-11
More colours for flamingo

bwvdnbro:
This MR adds more colours to the default FLAMINGO colour cycle to help avoid clashes. It also reorders the GSMF plots so that the 50 kpc aperture plots are shown first.


Moyoxkit (commented):
Looks good!

bwvdnbro:
Because one of the stellar mass function plots would show simulations and observations as a line with the same colour if you plotted more than one simulation:
image
This PR does not fix that issue, but it makes it less likely. COLIBRE already used more colours.

PR #174, opened 2022-04-13
COLIBRE CDDF plot

bwvdnbro:
Plot of the column density distribution function (CDDF) for COLIBRE.

Example:
image

This will also require the changes in the observational data from https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/112

### DIFF
--- a/colibre/scripts/cddf.py
+++ b/colibre/scripts/cddf.py
@@ -10,7 +10,6 @@ from astropy.cosmology import z_at_value
 import astropy.units as u
 
 import glob
-import os
 
 
 def plot_cddf(snapshot_filenames, names, output_path, observational_data):
@@ -28,7 +27,6 @@ def plot_cddf(snapshot_filenames, names, output_path, observational_data):
 
         # compute an appropriate number of pixels to use
         num_pix = int(4000.0 * (boxsize / (12.5 * Mpc)))
-        print(f"Using {num_pix}^2 pixels for CDDF projection")
 
         # create the HI mass fraction dataset
         data.gas.HI_mass_fraction = (
@@ -73,6 +71,7 @@ def plot_cddf(snapshot_filenames, names, output_path, observational_data):
             )
             - z
         )
+        # correct the CDDF for line of sight sampling effects
         dX = (
             (data.metadata.cosmology.H0 / data.metadata.cosmology.H(z))
             * dz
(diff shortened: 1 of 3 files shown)


colibre/scripts/cddf.py
```diff
@@ -0,0 +1,130 @@
...
+        data.gas.HI_mass_fraction = (
+            data.gas.masses
+            * data.gas.element_mass_fractions.hydrogen
+            * data.gas.species_fractions.HI
+        )
+        HI_map = project_gas(
+            data,
+            resolution=num_pix,
+            project="HI_mass_fraction",
+            parallel=True,
```
EvgeniiChaikin, line 41: This might be a surprise for someone who expects the pipeline to be run on a single thread.

Maybe set to `False` by default but pass a `bool` parameter to the script in the same way as in

https://github.com/SWIFTSIM/pipeline-configs/blob/master/colibre/config.yml#L221

so that the user can specify in the config file whether he/she wants to run this script in parallel?

@bwvdnbro what do you would be best?

EvgeniiChaikin (commented):
Hi @bwvdnbro 

Everything seems to work!

- I would suggest that we first merge https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/112, and then in this PR we update the pointer to the `velociraptor-comparison-data` repository so that the user can download the latest data including the data for the CDDF.

- What this PR is still missing is ~5 lines added to https://github.com/SWIFTSIM/pipeline-configs/blob/master/colibre/config.yml to include the new CDDF script.

- I would probably change the name of the script to `column_density_distribution_function `.

If you agree with the above suggestions, I can add them to the branch myself.

colibre/scripts/cddf.py
bwvdnbro: I actually don't know if doing this in parallel makes a huge difference. This was part of the original script. So I have no problems with not doing this in parallel. On the other hand, I can't think of many scenarios where you would want to make sure the pipeline is run on a single thread. Having more threads than idle cores is not really a problem, then it will just be less parallel than expected.

Having to change a parameter in the config file is about as easy as having to change this value in the script itself, so I'm not a huge fan of that.

bwvdnbro: Turns out that the speed difference is quite significant, even for a 6 Mpc box.

So what if we go with your suggestion, but make the default `parallel=True`?

EvgeniiChaikin: I think this sounds good!

PR #178, opened 2022-04-29
Added additional CDDF plot that only uses half of the box.

bwvdnbro:
The idea behind this is that low and intermediate column densities will be affected by a change in the integration depth. High column densities are usually completely dominated by a very small region and are less sensitive to this. As a result, comparing different box sizes makes little sense for lower column densities. Having a second plot where the integration depth is only half the box size can help to estimate how much impact the integration depth has on the CDDF.

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -286,8 +286,17 @@ scripts:
     section: Stellar Metallicity
     title: 'Fraction of Fe from SNIa vs [O/H]'
   - filename: scripts/column_density_distribution_function.py
+    title: Column Density Distribution Function (full box)
     caption: 'The column density distribution function of neutral hydrogen computed by projecting the entire box along the z-axis, corrected for the size of the box in redshift space.'
-    output_file: column_density_distribution_function.png
+    output_file: column_density_distribution_function_chunk1.png
     section: Column Densities
     additional_arguments:
       parallel: True
+  - filename: scripts/column_density_distribution_function.py
+    title: Column Density Distribution Function (half box)
+    caption: 'The column density distribution function of neutral hydrogen computed by projecting half the box along the z-axis, corrected for the size of the box in redshift space.'
+    output_file: column_density_distribution_function_chunk2.png
+    section: Column Densities
+    additional_arguments:
+      parallel: True
+      box_chunks: 2
--- a/colibre/scripts/column_density_distribution_function.py
+++ b/colibre/scripts/column_density_distribution_function.py
@@ -4,7 +4,7 @@ import matplotlib.pyplot as plt
 from swiftsimio.visualisation.projection import project_gas
 from velociraptor.observations import load_observation
 
-from unyt import Mpc, g, cm, mp
+import unyt
 
 from astropy.cosmology import z_at_value
 import astropy.units as u
@@ -13,7 +13,12 @@ import glob
 
 
 def plot_cddf(
-    snapshot_filenames, names, output_path, observational_data, parallel=True
+    snapshot_filenames,
+    names,
+    output_path,
+    observational_data,
+    box_chunks=1,
+    parallel=True,
 ):
     simulation_lines = []
     simulation_labels = []
@@ -28,7 +33,7 @@ def plot_cddf(
         boxsize = data.metadata.boxsize[0]
 
         # compute an appropriate number of pixels to use
-        num_pix = int(4000.0 * (boxsize / (12.5 * Mpc)))
+        num_pix = int(4000.0 * (boxsize / (12.5 * unyt.Mpc)))
 
         # create the HI mass fraction dataset
         data.gas.HI_mass_fraction = (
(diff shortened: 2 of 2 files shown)


colibre/config.yml
```diff
@@ -286,8 +286,17 @@ scripts:
...
+    title: Column Density Distribution Function (full box)
     caption: 'The column density distribution function of neutral hydrogen computed by projecting the entire box along the z-axis, corrected for the size of the box in redshift space.'
-    output_file: column_density_distribution_function.png
+    output_file: column_density_distribution_function_chunk1.png
     section: Column Densities
     additional_arguments:
       parallel: True
+  - filename: scripts/column_density_distribution_function.py
+    title: Column Density Distribution Function (half box)
+    caption: 'The column density distribution function of neutral hydrogen computed by projecting half the box along the z-axis, corrected for the size of the box in redshift space.'
```
EvgeniiChaikin, line 297:
```suggestion
    caption: 'The column density distribution function of neutral hydrogen computed by splitting the box in half along the z-axis and projecting each half separately, also along the z-axis. Corrected for the size of the box in redshift space.'
```

colibre/config.yml
```diff
@@ -286,8 +286,17 @@ scripts:
     section: Stellar Metallicity
     title: 'Fraction of Fe from SNIa vs [O/H]'
   - filename: scripts/column_density_distribution_function.py
+    title: Column Density Distribution Function (full box)
     caption: 'The column density distribution function of neutral hydrogen computed by projecting the entire box along the z-axis, corrected for the size of the box in redshift space.'
-    output_file: column_density_distribution_function.png
+    output_file: column_density_distribution_function_chunk1.png
     section: Column Densities
     additional_arguments:
       parallel: True
```
EvgeniiChaikin, line 294: Maybe pass the args in a bit more explicit way?

```suggestion
      parallel: True
      box_chunks: 1
```

EvgeniiChaikin (commented):
@bwvdnbro I have added just two small comments.

bwvdnbro:
Thanks for the review!

I have been thinking a bit more about these plots since Friday, especially now that Joop also suggested to add another plot with a reduced x-axis range. The problem is that these plots are relatively expensive compared to other pipeline plots. So we probably don't want to rerun the script multiple times, especially not if it is just to reduce the x-axis range. The pipeline has support for a single script that generates multiple figures, so maybe this would be a good place to use it. That is a no-brainer for the reduced axis range plot, but actually also works for the two box sizes: the full box map can be obtained by summing the two half box maps. So a single pass for each box half (the expensive part of the calculation) is sufficient to generate all of the plots.

What do you think?

EvgeniiChaikin:
Yes, you are right. The script is overly expensive and we can optimize it.

- Since `N_HI` is an integral over the box size, I presume what we need to do first is to compute

```
this_HI_map = (
                project_gas(
                    data,
                    resolution=num_pix,
                    project="HI_mass_fraction",
                    parallel=parallel,
                    backend="subsampled",
                    region=[
                        0.0 * boxsize,
                        boxsize,
                        0.0 * boxsize,
                        boxsize,
                        (1.0 / box_chunks) * i * boxsize,
                        (1.0 / box_chunks) * (i + 1) * boxsize,
                    ],
                )
```

for each chunk separately and without flattening at the end (to preserve the 2D shape). And then to get `N_HI` for the full box, we would need to add all these 2D maps together.

- If we want to do everything within a single script run (currently we run it twice), we could make the argument `box_chunks` be a list. So, for example, if we want `N_HI` for the full box and for the box split in haft, then we could pass `box_chunks: [1, 2]`. Within the script, to decide how many chunks we need in total to split the box into, we could use `N_chunks = np.max(box_chunks)` What do you think?

bwvdnbro:
I don't know how general we need to make the script. If we are happy to always only use the full and half box, then we could just hardcode that. If we decide to be more general, then we probably also want to pass on arguments that set the plot ranges.

I'm also not sure what the best way would be to generate the CDDF for multiple simulations in comparison mode. I don't think we want to store multiple maps for multiple simulations in memory. So we probably want to do a first loop over simulations where we compute all the CDDFs for a single simulation at a time, and then a second loop where we actually create the plots.

EvgeniiChaikin:
I agree that having a loop over the simulations and inside it, a loop over chucks is probably best.

I'd say that using _fixed_ plot ranges for all `N_HI` plots would make it easier to see the effect of the box size (by visually comparing the `N_HI` plots made for different sizes along the z-axis).

bwvdnbro:
I have implemented the new version we discussed. One caveat: there is currently no way to pass command line arguments with multiple values to the script, because `ScriptArgumentParser` does not support the `nargs="+"` argument of `argparse.ArgumentParser.add_argument()`. Implementing this is not very high on my to-do list.

For now, I am happy with this version; if we want to change the ranges or add more plots, we need to change the hard-coded values in the script, but it is quite clear where that needs to happen.

EvgeniiChaikin:
@bwvdnbro, everything works smoothly.

One last comment: how about we increase the upper y limit in the full-range plots a little bit? In one of my tests of your branch, the simulation data is above the current y upper limit

column_density_distribution_function_chunk2_full_range

Otherwise, the branch should be ready for merging.

bwvdnbro:
Done. My test simulation did not probe the low density regime well, so I didn't see this problem.

bwvdnbro:
Since you asked yesterday, for reference: the dX correction factor can be found e.g. in [Berg et al. (2019)](https://ui.adsabs.harvard.edu/abs/2019MNRAS.488.4356B/abstract), eq. (3):
image

The (1+z)^2 is part of that factor and is not the same as the conversion factor from co-moving to physical number densities that is applied earlier in the script.

EvgeniiChaikin:
I see. Thank you for pointing out where that expression comes from!

PR #179, opened 2022-05-11
Added FLAMINGO HSE data to COLIBRE gas fraction plots.

bwvdnbro:
The FLAMINGO data were corrected for hydrostatic bias using the FLAMINGO correction factor. They include some of the other datasets and will be useful when comparing different jet models in the future.


EvgeniiChaikin (commented):
Looks good!

gas_fraction_500

PR #180, opened 2022-05-19
New dust plots multisize

james-trayford:
See also https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/114. Incorporate new plots for multi-size dust runs.

Some of these require several new VR properties, which arise from new COLIBRE pre-processing fields. I have the VR config for this, should I also incorporate this into the PR?


bwvdnbro:
Yes, it is probably a good idea to also update the VR config file. I suspect we have not always done that in the past, because the current one is 15 months old. I assume you will also update the VR config file in the SWIFT repository (if you haven't done that already)?

james-trayford:
> Yes, it is probably a good idea to also update the VR config file. I suspect we have not always done that in the past, because the current one is 15 months old. I assume you will also update the VR config file in the SWIFT repository (if you haven't done that already)?

Yes I should update that also - the reason I hesitate to update the pipeline VR config is that while the pipeline + velociraptor python handle missing fields OK, running VR with the a config asking for fields that aren't present will crash, and some of these fields are specific to the new branch, so will break if run on older code / other branches of the code. 

Perhaps it does make sense for the VR config file to be kept up to date in the `SWIFT-COLIBRE` branch, and to keep a minimal example in the `pipeline-configs` (at least until the new generated fields are more established)?

bwvdnbro:
I agree that we should make sure that we can still run VR on older snapshots until the new dust model is fully merged in SWIFT. Keeping the VR config file in the SWIFT repository up to date then makes sense, since that is immediately linked to a specific code version. I'm actually not quite sure why we also have it here, since that can only lead to confusion.

colibre/auto_plotter/grain_size_comparison.yml
```diff
@@ -0,0 +1,175 @@
...
+    number_of_bins: 15
+    start:
+      value: 3e7
+      units: Solar_Mass
+    end:
+      value: 3e12
+      units: Solar_Mass
+  metadata:
+    title: Stellar mass vs Small-to-Large Ratio
+    caption: The dust mass ratio of small to large grains in all gas averaged of 50 kpc in galaxies. Compared to values derived from observation by Relano et al 2020.
```
bwvdnbro, line 29: I'm not sure I understand the "averaged of 50 kpc" bit of the caption. Is this the average over a 50 kpc aperture?

colibre/registration.py
```diff
@@ -163,37 +159,322 @@ def register_dust(self, catalogue, aperture_sizes):
...
+            catalogue.lin_element_ratios_times_masses,
+            f"lin_O_over_H_total_times_gas_mass_100_kpc",
+        )
+
+        logOH_abundance_times_mhi = (
+            getattr(
+                catalogue.log_element_ratios_times_masses,
+                f"log_O_over_H_atomic_times_gas_mass_lowfloor_{aperture_size}_kpc",
+            )
+            / 0.7
```
bwvdnbro, line 249: Could you add a comment that explains the origin of the `0.7` factor?

colibre/scripts/density_species_depletions.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+
+    data = load(filename)
+    pairing = int(data.metadata.parameters["DustEvolution:pair_to_cooling"])
+
+    if int(data.metadata.parameters["DustEvolution:silicate_fe_grain_fraction"]) != -1:
+        fe_balance = float(
+            data.metadata.parameters["DustEvolution:silicate_fe_grain_fraction"]
+        )
+        mol_O = 6 * A["Oxygen"]
```
bwvdnbro, line 74: Could you add a small comment explaining where these factors come from?

colibre/scripts/density_species_depletions.py
```diff
@@ -0,0 +1,402 @@
...
+    mask = H_norm == 0.0
+    Hd[mask] = None
+    Hh2[mask] = None
+    Hhi[mask] = None
+    Hhii[mask] = None
+    H_h[mask] = 1.0
+    H_Z[mask] = 1.0
+    H_norm[mask] = 1.0
+
+    # construct dictinary of arrays for individual dust species
```
bwvdnbro, line 223:
```suggestion
    # construct dictionary of arrays for individual dust species
```

colibre/scripts/density_species_diffuse.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+
+    data = load(filename)
+    pairing = int(data.metadata.parameters["DustEvolution:pair_to_cooling"])
+
+    if int(data.metadata.parameters["DustEvolution:silicate_fe_grain_fraction"]) != -1:
+        fe_balance = float(
+            data.metadata.parameters["DustEvolution:silicate_fe_grain_fraction"]
+        )
+        mol_O = 6 * A["Oxygen"]
```
bwvdnbro, line 74: Same comment as for the other script

colibre/scripts/density_species_diffuse.py
```diff
@@ -0,0 +1,404 @@
...
+    mask = H_norm == 0.0
+    Hd[mask] = None
+    Hh2[mask] = None
+    Hhi[mask] = None
+    Hhii[mask] = None
+    H_h[mask] = 1.0
+    H_Z[mask] = 1.0
+    H_norm[mask] = 1.0
+
+    # construct dictinary of arrays for individual dust species
```
bwvdnbro, line 226:
```suggestion
    # construct dictionary of arrays for individual dust species
```

bwvdnbro (commented):
I have added some small comments.

Do we want to merge this now, or do we wait until the dust model has been merged in the code as well?

MatthieuSchaller:
The dust model is now integrated in the main code so if the config still needs updating, let me know.

colibre/auto_plotter/grain_size_comparison.yml
james-trayford: its just the mass ratio aggregated over the aperture, so 'average' isn't really accurate here, changed to reflect

colibre/scripts/density_species_diffuse.py
james-trayford: added

colibre/registration.py
james-trayford: thanks, there's not a good reason for these, think this is a hangover from debugging

james-trayford:
@MatthieuSchaller I think this is ready - I've implemented @bwvdnbro 's suggestions here and in the VRpy repo (https://github.com/SWIFTSIM/velociraptor-python/pull/84) - I've also added the requisite VR config updates to a COLIBRE in an MR [here](https://gitlab.cosma.dur.ac.uk/EAGLE/swift-colibre/-/merge_requests/236)

MatthieuSchaller:
Great, thanks!

PR #182, opened 2022-05-23
New FLAMINGO-HSE gas fraction plot and updated caption

bwvdnbro:


Moyoxkit:
Looks good, this plot already exist in the "calibration" section, but it will be good to also have it with the other gas fraction plots

bwvdnbro:
To be clear: this is for COLIBRE, which has no calibration section. And I'm pretty sure the plot is already included in the gas fractions section for FLAMINGO.

Is the new caption accurate?

Moyoxkit:
Missed the colibre part, the captions are correct

PR #184, opened 2022-06-01
Add luminosity functions to colibre

MatthieuSchaller:
Also update the data submodule to the latest version.


colibre/registration.py
```diff
@@ -901,3 +927,4 @@ def register_stellar_mass_scatter(self, catalogue, stellar_mass_scatter_amplitud
 register_star_Mg_and_O_to_Fe(self, catalogue, aperture_sizes_30_50_100_kpc)
 register_gas_fraction(self, catalogue)
 register_stellar_mass_scatter(self, catalogue, stellar_mass_scatter_amplitude)
+register_star_magnitudes(self, catalogue, aperture_sizes)
```
EvgeniiChaikin, line 930:
```suggestion
register_star_magnitudes(self, catalogue, aperture_sizes_30_50_100_kpc)
```

EvgeniiChaikin (commented):
Thanks @MatthieuSchaller 

One comment - don't we want to use the 50 kpc apertures, like in the other plots?

colibre/registration.py
MatthieuSchaller: Yes, good point. 30kpc was the eagle choice. I should have changed it.

Can you make the change here? I am off laptop today.

EvgeniiChaikin: Sure, done

MatthieuSchaller:
Thanks!

PR #189, opened 2022-06-24
Draft: Added example box size correction.

bwvdnbro:
To be combined with the same branch in the velociraptor-python and pipeline repositories.


colibre/auto_plotter/mass_functions.yml
```diff
@@ -116,6 +116,7 @@ stellar_mass_function_50:
   type: "massfunction"
```
EvgeniiChaikin, line 116: Note that here the type of the plot is `massfunction` but in `correction.yml`, it is `mass_function`. Shouldn't these two be the same?

bwvdnbro: No, because they are actually not the same thing. `massfunction` is a plot type, while `mass_function` is a line type. That is what the autoplotter uses internally. I don't know if that distinction is useful.

If we would use `massfunction`, then the `velociraptor-compute-box-size-correction` script would not be able to find the line in the pipeline `data*.yml` output file. I could hardcode a conversion `mass_function-->massfunction` into the script if you think that is useful.

bwvdnbro:
To be replaced with a new PR that includes a more appropriate correction file and adds an additional plot showing the correction.

bwvdnbro:
Replaced by #197.

PR #191, opened 2022-06-28
Output CDDF plot metadata in yml file.

bwvdnbro:
This outputs the data for the various CDDF plots in a .yml file for each simulation processed, as is done for the autoplotter plots.

@EvgeniiChaikin I think this should be compatible with your new hypercube plotting tool.


EvgeniiChaikin (commented):
@bwvdnbro All looks good, I think this is exactly what we need right now! I left just one comment in the code diff.

Maybe at some point in the future we could also consider storing more metadata in these yaml files, such as

-  box_chunks
-  x_ranges
-  y_ranges
-  redshifts of the simulations
-  names of the simulations

in order to always know exactly which simulation and redshift the data comes from and for what script's setting.

bwvdnbro:
I agree that more metadata would be a good idea. But I also think it would be better to have a more general way of producing these .yml files for plots that do not use the autoplotter, so maybe we can find a better way to do both of these in the future.

PR #194, opened 2022-07-18
Update abundance plots

correac:
New set of plots for metallicity and stellar abundances (have a look here https://swift.dur.ac.uk/COLIBREPlots/ccorrea/pipeline/2022_06_comparison/PipelineUpdate/)


bwvdnbro:
Could you remove the `__pycache__` and `DS_store` files from the pull request? And I assume `config2.yml` is also a leftover from debugging? Please also run the formatting script.

correac:
Oh right, forgot to clean up!

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
```diff
@@ -0,0 +1,157 @@
...
+from unyt import unyt_array
+
+
+def read_data(data):
+    """
+    Grabs the data
+    """
+
+    mH_in_cgs = unyt.mh
+    mFe_in_cgs = 55.845 * unyt.mp
```
bwvdnbro, line 22: Sorry for being so pedantic, but do you have a reference for these values?

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
```diff
@@ -0,0 +1,157 @@
...
+    Fe_H, C_Fe = read_data(data)
+
+    # Bins along the X axis (Fe_H) to plot the median line
+    bins = np.arange(-4.1, 1, 0.2)
+    ind = np.digitize(Fe_H, bins)
+
+    xm, ym, ym1, ym2 = [], [], [], []
+    Min_N_points_per_bin = 11
+
+    for i in range(1, len(bins)):
```
bwvdnbro, line 83: I have some comments about this binning strategy.

I'm not quite sure it is correct to use the median for both the x and y variable in a bin. When you bin quantities like this, then the x variable is treated as an independent variable for which you impose a value (the bin centre) and an associated uncertainty (the bin width). The y variable is then a dependent variable that has a median and uncertainty given those values of the independent variable. By using the median for x as well, you arbitrarily shift the x values within their uncertainty interval (the bin), without really adding any useful information.

Manually binning values like this is also really slow compared to using `scipy.stats.binned_statistic`. You can get the same curve (without medians in x) using
```
xm = 0.5 * (bins[1:] + bins[:-1])
ym,_,_ = stats.binned_statistic(Fe_H, C_Fe, statistic="median", bins=bins)
ym1,_,_ = stats.binned_statistic(Fe_H, C_Fe, statistic=lambda x: np.percentile(x, 16.), bins=bins)
ym2,_,_ = stats.binned_statistic(Fe_H, C_Fe, statistic=lambda x: np.percentile(x, 84.), bins=bins)
counts,_,_ = stats.binned_statistic(Fe_H, C_Fe, statistic="count", bins=bins)
mask = counts >= Min_N_points_per_bin
xm = xm[mask]
ym = ym[mask]
ym1 = ym1[mask]
ym2 = ym2[mask]
```

bwvdnbro (changes requested):
I have left extensive comments in one of the new scripts, but they also apply to most of the other scripts.

Could you please remove `eagle-xl/scripts/__pycache__/load_sfh_data.cpython-36.pyc` as well?

It looks like `Buder21_data.hdf5` does not actually comply with our usual data format after all. Could you move it to `StellarAbundances/raw`? For now, it would still be okay to use it directly from that folder and then at least it is clear this file cannot be used in the autoplotter.

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
correac: I do not have reference for those. Regarding Buder21_data.hdf5, I did not compile that dataset, James Trayford did. So let's leave it in StellarAbundances until James can add the Buder21 data treatment in the raw/ folder.

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
```diff
@@ -0,0 +1,157 @@
...
+            zorder=1000,
+            path_effects=[pe.Stroke(linewidth=4, foreground="white"), pe.Normal()],
+        )[0]
+    )
+    simulation_labels.append(f"{name} ($z={redshift:.1f}$)")
+
+# We select all APOGEE files containing FeH-CFe.
+path_to_obs_data = f"{arguments.config.config_directory}/{arguments.config.observational_data_directory}"
+observational_data = f"{path_to_obs_data}/data/StellarAbundances/APOGEE_data_C.hdf5"
+x = unyt_array.from_hdf5(observational_data, dataset_name="values", group_name="x")
```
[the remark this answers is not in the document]
correac: I tried but it gave me errors.

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
correac: I followed the median binning, since this was how it was originally implemented in stellar_abundances_FeH_MgFe.py in the main branch of the pipeline-config/colibre. But I can update it to use scipy.

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Create a histogram
+h, xedges, yedges = np.histogram2d(x.value, y.value, bins=(xi, yi))
+xbins = xedges[:-1] + (xedges[1] - xedges[0]) / 2
+ybins = yedges[:-1] + (yedges[1] - yedges[0]) / 2
+
+z = h.T
+
+binsize = 0.25
+grid_min = np.log10(10)
```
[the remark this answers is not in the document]
correac: Yes, I was playing with the min grid and I ended up using 10, which is 1 indeed. :)

colibre/scripts/stellar_abundances_FeH_CFe_APOGEE.py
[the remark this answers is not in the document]
bwvdnbro: What kind of errors?

colibre/scripts/stellar_abundances_FeH_OFe_GALAH.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    simulation_labels.append(f"{name} ($z={redshift:.1f}$)")
+
+# We select all GALAH file containing FeH-MgFe.
+path_to_obs_data = f"{arguments.config.config_directory}/{arguments.config.observational_data_directory}"
+observational_data = f"{path_to_obs_data}/data/StellarAbundances/Buder21_data.hdf5"
+
+GALAH_data = h5py.File(observational_data, "r")
+galah_edges = np.array(GALAH_data["abundance_bin_edges"])
+obs_plane = np.array(GALAH_data["Mg_enrichment_vs_Fe_abundance"]).T
```
bwvdnbro, line 118:
```suggestion
obs_plane = np.array(GALAH_data["O_enrichment_vs_Fe_abundance"]).T
```

bwvdnbro:
For reference: after looking carefully at things, I realised there was quite a lot of code duplication going on in the various scripts. I have eliminated duplicate code by combining all scripts into a single script that takes additional arguments. The plots have not changed (except for some small things, like location of the legend).

Also, the data shown in one of the GALAH plots was wrong: instead of showing [O/Fe], this plot was showing the [Mg/Fe] data again. The correct plot looks like this:
image

I have also updated the reference to the data repository.

Because I did not have write access to the original branch for this PR, this was merged manually into master.

PR #196, opened 2022-07-27
Added percentage of particles included in distribution to metal mass fraction distribution plot

bwvdnbro:
Example of the new plot:
metallicity_distribution


EvgeniiChaikin (commented):
This all looks good, @bwvdnbro!

Sorry that it took me a while to have a look into this PR.

Note that this branch needs to be rebased on the most recent master before being merged.

PR #200, opened 2022-08-03
Add high z eagle gsmf to the COLIBRE pipeline

MatthieuSchaller:
And update the observational data submodule to the lastest master.


EvgeniiChaikin:
I think we should store EAGLE GSMF data in a **multi-z** catalogue because we have this data at several redshifts. Then in a yml file, we will not need to specify a redshift at which to plot the data, and we won't have EAGLE GSMF appearing at z=0.0, 0.1, and z=0.5 in the same plot.

I created a PR there https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/129. After that PR is merged, I can update this branch.

EvgeniiChaikin (commented):
I updated the branch as explained in the above message. I think it is now ready to be merged.

PR #203, opened 2022-08-09
Added stellar birth pressure distribution plot.

bwvdnbro:
This adds a new plotting script that creates a stellar birth pressure distribution plot, similar to the existing stellar birth density distribution plot. The stellar birth pressure is simply determined by multiplying the stellar birth number density with the stellar birth temperature.

Example plot:
birth_pressure_distribution


EvgeniiChaikin (commented):
@bwvdnbro I ran this branch on several simulations and I confirm that everything works as expected!

I don't have any comments about the new code (I can see that the stellar birth pressure script is essentially the same as the stellar birth density script except that density is replaced with pressure)

PR #207, opened 2022-08-18
Fixed layout issues in multi-panel comparison plots.

bwvdnbro:
When run in comparison mode, a lot of scripts produce multi-panel plot. If the run names are long, the layout of this plot usually looks pretty awful, e.g.:
density_temperature_old

I fixed this by (a) reducing the font size of the name, and (b) by making sure the name is not taken into account when applying `tight_layout`. Normal length names will now show up properly, and unusually long names will simply be hidden below the other axes.

Result:
density_temperature_new


EvgeniiChaikin (approved):
Thanks, @bwvdnbro! The script-made plots with decreased font in their legends look much better after this update in comparison mode!

PR #210, opened 2022-08-22
Fixed some SOAP specific things in the FLAMINGO registration functions.

bwvdnbro:
This is still work in progress.

Fixes some problems with the FLAMINGO registration functions when running the pipeline on the SOAP catalogue output instead of the VR output:
 - SOAP has metal masses instead of mass fractions, so we still need to divide out the total mass
 - SOAP uses different units for the SFR, which led to an interesting bug when computing the SSFR, because of the known issue that `unyt` array assignment ignores units completely.

### DIFF
--- a/flamingo/registration.py
+++ b/flamingo/registration.py
@@ -46,7 +46,6 @@ for aperture_size in aperture_sizes:
     ssfr[good_stellar_mass] = (
         star_formation_rate[good_stellar_mass] / stellar_mass[good_stellar_mass]
     ).to(marginal_ssfr.units)
-    print(ssfr)
 
     # Name (label) of the derived field
     ssfr.name = f"Specific SFR ({aperture_size} kpc)"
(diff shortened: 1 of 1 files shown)


flamingo/registration.py
```diff
@@ -42,9 +42,11 @@
     star_formation_rate = getattr(catalogue.apertures, f"sfr_gas_{aperture_size}_kpc")
 
     ssfr = np.ones(len(star_formation_rate)) * marginal_ssfr
+    # be aware of units here! unyt array assignment does not check units!
     ssfr[good_stellar_mass] = (
         star_formation_rate[good_stellar_mass] / stellar_mass[good_stellar_mass]
-    )
+    ).to(marginal_ssfr.units)
+    print(ssfr)
```
Moyoxkit, line 49: You left a debug print statement

bwvdnbro: Good point. I took it out.

PR #211, opened 2022-08-24
Added 2 extra colours to the default colour cycle.

bwvdnbro:
After seeing these plots today:
image
I decided it is time to add some more colours to the default colour cycle. Then we get:
image
The extra colours are up for discussion, but I noticed that we do not have any shade of yellow yet, or any dark green, so that is why I picked these.


EvgeniiChaikin (approved):
Looks good!

PR #214, opened 2022-09-14
Add spin/jet pipeline for colibre

FHusko:
Identical to the flamingo one, but the scatter plots are turned on by default, since the number of objects is much smaller.

Here is an example of the output (some of the non-BH plots are incorrect for unrelated reasons, please ignore those): [http://icc.dur.ac.uk/~xqzq36/COLIBRE_50/colibre_example/](http://icc.dur.ac.uk/~xqzq36/COLIBRE_50/colibre_example/).

PR #219, opened 2022-10-05
Add five black-hole-related scripts/plots

EvgeniiChaikin:
This PR adds five black-hole-related scripts/plots from @FHusko's jets update (https://github.com/SWIFTSIM/pipeline-configs/pull/214)

Namely, the scripts are

- scripts/bh_eddington_fractions.py
- scripts/bh_merger_mass_fractions.py
- scripts/bh_luminosities.py
- scripts/bh_thermal_energies.py
- scripts/bh_thermal_heatings.py

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -167,6 +167,31 @@ scripts:
     output_file: bh_masses.png
     section: Black Holes
     title: Black Hole Dynanmical and Subgrid Masses
+  - filename: scripts/bh_eddington_fractions.py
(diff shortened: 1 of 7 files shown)


EvgeniiChaikin:
Note that the scripts here are not fully equivalent to those in #214 because I also

- lowered the min BH mass (since COLIBRE uses higher resolution/lower seed mass than FLAMINGO)
- removed library imports that are not used in the scripts
- replaced general except statements with more specific ones.

bwvdnbro (approved):

PR #220, opened 2022-10-07
Add black hole mass function plots

EvgeniiChaikin:


FHusko:
Perhaps it would be good to also change these to start from 1e3 Msol for the BH masses?

bwvdnbro:
Since I will be on holiday for the next 2 weeks, I have assigned this PR to @MatthieuSchaller 

I can't seem to assign @FHusko as reviewer, so I will just approve this now on the condition that his comments are addressed. Hopefully this can then be merged without my help.

bwvdnbro (approved):

EvgeniiChaikin:
@FHusko Yes, indeed! It is always good to plot the whole available range. I have just committed this change.

Here is how the new plot looks like
 
black_hole_mass_function

FHusko:
@EvgeniiChaikin Great, thanks! Looking good.

PR #223, opened 2022-10-20
Update the SNIa plots to include the observations

Fonotec:
These are the current update to the plots of the SNIa rates.

### DIFF
--- a/colibre/registration.py
+++ b/colibre/registration.py
@@ -96,6 +96,18 @@ def register_spesific_star_formation_rates(self, catalogue, aperture_sizes):
         )
         is_bigger_than_1e10.name = "Stellar mass larger than 10^10 Msun"
 
+        # Mask for galaxies above 10^10 Msun
+        is_bigger_than_1e10_active = unyt.unyt_array(
+            (
+                (stellar_mass > unyt.unyt_quantity(1e10, units="Msun"))
+                & (ssfr > 1.01 * marginal_ssfr)
+            ).astype(float),
+            units="dimensionless",
+        )
+        is_bigger_than_1e10_active.name = (
+            "Stellar mass larger than 10^10 Msun and active"
+        )
+
         # Mask for galaxies above 5* 10^10 Msun
         is_bigger_than_5e10 = unyt.unyt_array(
             (stellar_mass > unyt.unyt_quantity(5e10, units="Msun")).astype(float),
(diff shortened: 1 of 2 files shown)


MatthieuSchaller:
Assigned to Bert as he is already looking at the data side.

colibre/registration.py
```diff
@@ -94,7 +94,38 @@ def register_spesific_star_formation_rates(self, catalogue, aperture_sizes):
             (stellar_mass > unyt.unyt_quantity(1e10, units="Msun")).astype(float),
             units="dimensionless",
         )
-        is_active.name = "Stellar mass larger than 10^10 Msun"
+        is_bigger_than_1e10.name = "Stellar mass larger than 10^10 Msun"
+
+        # Mask for galaxies above 10^10 Msun
+        is_bigger_than_1e10_active = unyt.unyt_array(
+            (
+                (stellar_mass > unyt.unyt_quantity(1e10, units="Msun"))
```
bwvdnbro, line 102: Some recent experiences have taught us that you cannot rely on `unyt` doing the right thing when comparing `unyt_array`s; it tends to just completely ignore units. I therefore propose the following:
```
stellar_mass > (1.e10*unyt.Msun).to(stellar_mass.units)
...
ssfr > (1.01*marginal_ssfr).to(ssfr.units)
```

colibre/registration.py
```diff
@@ -94,7 +94,38 @@ def register_spesific_star_formation_rates(self, catalogue, aperture_sizes):
...
+        is_bigger_than_5e10 = unyt.unyt_array(
+            (stellar_mass > unyt.unyt_quantity(5e10, units="Msun")).astype(float),
+            units="dimensionless",
+        )
+        is_bigger_than_5e10.name = "Stellar mass larger than 5 $\\times$ 10^10 Msun"
+
+        # Mask for galaxies above 5 * 10^10 Msun
+        is_bigger_than_5e10_active = unyt.unyt_array(
+            (
+                (stellar_mass > unyt.unyt_quantity(5e10, units="Msun"))
```
bwvdnbro, line 121: Same here.

colibre/registration.py
```diff
@@ -94,7 +94,38 @@ def register_spesific_star_formation_rates(self, catalogue, aperture_sizes):
...
+            ).astype(float),
+            units="dimensionless",
+        )
+        is_bigger_than_1e10_active.name = (
+            "Stellar mass larger than 10^10 Msun and active"
+        )
+
+        # Mask for galaxies above 5* 10^10 Msun
+        is_bigger_than_5e10 = unyt.unyt_array(
+            (stellar_mass > unyt.unyt_quantity(5e10, units="Msun")).astype(float),
```
bwvdnbro, line 113: Same here.

bwvdnbro (changes requested):
I have added a comment to the registration functions. This issue recently popped up when using the SOAP rather than VR halo catalogue, since the units for the SFR are different there. Apparently `unyt` just ignores the units in comparisons, so if you use the wrong units, you get the wrong mask. I realise that some other registration functions are probably also wrong in that sense, but those can be fixed when we actually switch to using SOAP (that is still the plan).

I imagine that Joop will want more extensive captions for the new figures, but that can be done later too.

The link to the observational data repository still needs to be updated, but that can only be done after the changes there have been merged.

Fonotec:
This is with your suggestions, btw we talked about this earlier, the Wiseman data will not appear on the plots this way. It has an average of z=0.5 but I would like to have it on the plot by default without needing to change the redshift of the data myself. The data for these SNIa surveys also do not work like other galaxy observations because it spans a more extended range in redshifts.

bwvdnbro (approved):

Fonotec:
Okay, I tried to update the observational_data folder, let me know if I did this correct

bwvdnbro:
No, something must have gone wrong, because there is a conflict. I will create a new branch based on your changes and merge it manually.

bwvdnbro:
I have made the changes in #226. Turns out you need to make sure the branch is up to date with master before updating the submodule (I think the submodule version changed in master as well, hence the conflict).

PR #226, opened 2022-11-01
Update the SNIa plots to include the observations (second attempt)

bwvdnbro:


MatthieuSchaller (approved):
Same changes as the other branches so I'll approve.

PR #234, opened 2022-12-07
Median temperature vs density plot (with metallicity bins)

bwvdnbro:
This PR adds a plot of the median temperature as a function of density, for all gas and for gas in selected metallicity bins.

The plot looks like this:
median_temperature_density

Same, but for 2 and 3 snapshots being plotted:
median_temperature_density
median_temperature_density

As you can see, the legend is distributed over the panels if there are multiple plots, and is moved to the white area if that exists.

### DIFF
--- a/colibre/scripts/median_temperature_density.py
+++ b/colibre/scripts/median_temperature_density.py
@@ -98,7 +98,7 @@ if __name__ == "__main__":
                 nHmid, Tmed, label=label, color=line_properties["colors"][0]
             )
 
-            for iZ, (Zmin, Zmax) in enumerate(Zbounds):
+            for iZ, (Zmin, Zmax) in enumerate(Zbounds, start=1):
                 mask = np.logical_and(
                     metal_mass_fraction >= Zmin * solar_metal_mass_fraction,
                     metal_mass_fraction < Zmax * solar_metal_mass_fraction,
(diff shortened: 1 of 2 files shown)


EvgeniiChaikin (changes requested):
Thank you @bwvdnbro!

I left a few small suggestions in the script.

I also tested your script for a 5-run comparison. It looks good!
median_temperature_density

One thing I do not understand is why the blue curve (all gas) does not converge to the orange curve (high-Z gas) at high densities when the other curves (lower-Z gas) are not present indicating that there is only high-Z gas. I would naively expect the blue to be very close to the orange in this case.

bwvdnbro:
The different metallicity bins do not completely span the entire possible range of metallicity values, so maybe that is why it does not converge?

And of course, there is also the issue that lines only show up if you have at least two adjacent bins that are not NaN. So maybe the high density bins are too sparsely populated when the metallicity cuts are applied?

colibre/scripts/median_temperature_density.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+
+            for iZ, (Zmin, Zmax) in enumerate(Zbounds, start=1):
+                mask = np.logical_and(
+                    metal_mass_fraction >= Zmin * solar_metal_mass_fraction,
+                    metal_mass_fraction < Zmax * solar_metal_mass_fraction,
+                )
+                if mask.sum() > 0:
+                    Tmed, _, _ = stats.binned_statistic(
+                        nH[mask], T[mask], statistic="mean", bins=nHbin
```
EvgeniiChaikin, line 108:
```suggestion
                        nH[mask], T[mask], statistic="median", bins=nHbin
```

EvgeniiChaikin (approved):
@bwvdnbro I found the reasons why the curves would not converge. That's because the temperature of all gas was computed as `median`, whereas the temperature in a given metallicity range was computed as `mean`.

If you replace `mean` with `median` by accepting my suggestion below, you will get the following plot
median_temperature_density

which shows that two curves converge to the same values both at low Z and at high Z, as expected.

Apart from the above comment, this PR looks ready to be merged, so I will approve it in advance.

colibre/scripts/median_temperature_density.py
bwvdnbro: Oops, thanks for spotting that!

PR #236, opened 2022-12-14
Plot the metallicity distribution as a PDF rather than a number of particles.

bwvdnbro:


EvgeniiChaikin (approved):
@bwvdnbro I'd probably rename `metallicities = {...}` to  `fraction_of_particles = {...}`

Otherwise, everything looks perfect!

PR #238, opened 2023-01-25
Add galaxy stellar mass function plots showing only calibration data

EvgeniiChaikin:
**Motivation for this PR:**
In the current GSMF plots, we have a great deal of comparison data. It is very useful to have all of them in order to better know and see where we are compared to other research groups. 

However, for calibration purposes, we only need the data we calibrate to. Having too many different curves in the same plot makes it more difficult for us to see how close we are relative to the calibration data.

After this PR, we will have two sets of GSMF plots: one with many curves showing comparison data, and the other showing only the calibration data (Driver 2021 for z \approx 0, and Leja for high z).

stellar_mass_function_50
stellar_mass_function_50_calibration


MatthieuSchaller (approved):
All good for me though I'll continue to disagree that we should use the Leja data...

PR #239, opened 2023-02-22
Use new trujillo sizes

MatthieuSchaller:
Update the plotting scripts to use the erratum-corrected Trujillo sizes.


EvgeniiChaikin (approved):
All looks good!

**Before the change**

before


**After the change**

after

PR #245, opened 2023-04-13
Add plots showing H2 and HI gas mass versus stellar and halo mass

EvgeniiChaikin:
The four new plots can be found here: https://swift.dur.ac.uk/COLIBREPlots/chaikin/individual_runs/YEAR2022/z0.0_U112f_L12N376_U93_npivot0p3_Ppivot8e3_Mseed5e2_dTAGN9p5/index.html#8211386328891335256

For the new plots, I created a new section on the webpabe, which is called **cold gas masses**.

We need this update because looking only at the cold gas _fractions_ does not always tell the whole story.


MatthieuSchaller (approved):
All good for me!

PR #246, opened 2023-04-16
Fix three bugs in the coversion scripts of raw cosmic SFRD data

EvgeniiChaikin:
The raw data files with the data from Magnelli (2013),  Gruppioni (2013), and Schenker (2013), first provide redshifts (column 1), then y (column 2), then positive errors on y (column 3), and finally negative errors on y (column 4).

Due to a typo, in the conversion script, the positive errors are read as the negative, and the negative as the positive. This leads to the scatter along y whose values are all negative. The newer versions of matplotlib do not allow negative scatter; this is how I spotted this bug.

I know that after this update the errors will be read correctly because the raw files also include headers with the description, so it is clear which errors are possible and which are negative.


EvgeniiChaikin:
Here are the links to the raw data

- https://github.com/SWIFTSIM/velociraptor-comparison-data/blob/master/data/StarFormationRateHistory/raw/sfr_gruppioni2013.dat
- https://github.com/SWIFTSIM/velociraptor-comparison-data/blob/master/data/StarFormationRateHistory/raw/sfr_magnelli2013.dat
- https://github.com/SWIFTSIM/velociraptor-comparison-data/blob/master/data/StarFormationRateHistory/raw/sfr_schenker2013.dat

MatthieuSchaller (commented):
Thanks. Can the same changes be applied to the EAGLE and FLAMINGO flavours?

EvgeniiChaikin:
Done! Thanks for the feedback!

I also rebased on the updated version of the master branch; otherwise, there would be a merge conflict because I edited the same code in another pull request.

PR #248, opened 2023-04-16
Update comparison data in the cosmic star formation rate density plot

EvgeniiChaikin:
This follows Joop's request on slack to

- change the label for Gruppioni+ 20 from "IR" to "FIR"
- omit all data from before 2016 except Bouwens+ 15

The updated plot will look as follows

star_formation_history


MatthieuSchaller (approved):

MatthieuSchaller:
Do we want to keep all the commented-out code then?

MatthieuSchaller:
Ok, you convinced me.

PR #250, opened 2023-04-24
fix dust scripts to work more generically across dust models

james-trayford:
This is intended to fix the density-species-depletion and density-species-diffuse scripts to work correctly with current dust grain model (and more generically across previous and prospective future models)


MatthieuSchaller (approved):
Great, thanks. Much cleaner indeed to work with the snapshot meta-content directly.

PR #252, opened 2023-04-26
Add observational data for baryon factions and stellar fractions

EvgeniiChaikin:
Examples of the new data:

star_fraction_500
baryon_fraction_500


MatthieuSchaller:
Is the Zhang data on the second plot actually useful.... ?

EvgeniiChaikin:
As far as I know, these are the datasets that are used in the Flamingo papers and @Moyoxkit copied them here; so, to answer your question, yes.

MatthieuSchaller (approved):

PR #253, opened 2023-05-16
BCG stellar vs halo mass, BH mass vs. stellar mass, BH mass vel disp.

FHusko:
This PR makes the following three changes:

1. The stellar to halo mass relation plots for BCGs are created using recently added data.
2. The black hole - stellar mass relation for active and passive galaxies (from the simulations) are now plotted for centrals only, because the observational data by Terrazas et al. (2017) only includes what they consider centrals. This was agreed by Joop.
3. Added data to the BH mass vs. stellar LOS velocity dispersion plot (Saglia et al. 2016), also using recently added observational data.


MatthieuSchaller (approved):
That is fine by me but it is changing existing plots so I'll wait for Evgenii's approval too.

colibre/auto_plotter/black_holes.yml
```diff
@@ -293,10 +298,11 @@ stellar_veldisp_black_hole_mass_30:
       units: km/s
   metadata:
     title: LOS Stellar Velocity Dispersion-Black Hole Mass relation (30 kpc)
-    caption: The 3D stellar velocity dispersion is converted into a LOS velocity dispersion using a $1/\sqrt{3}$ correction factor. Note that the stellar velocity dispersion in observations is measured in a fixed 1 kpc aperture.
+    caption: The 3D stellar velocity dispersion is converted into a LOS velocity dispersion using a $1/\sqrt{3}$ correction factor. The stellar velocity dispersion from Sahu et al. (2016) was measured using a fixed 1 kpc aperture, while that from Saglia et al. (2016) was measured using the half-light radius aperture.
     section: Black Holes
     show_on_webpage: false
   observational_data:
+    - filename: StellarVelocityDispersionBlackHoleMass/Saglia2016.hdf5
```
EvgeniiChaikin, line 305: I don't think the version of the submodule `velociraptor-comparison-data` in this repository has been updated recently. This means that the current version of `velociraptor-comparison-data` used here does not include the `Saglia+2016` data.

colibre/auto_plotter/stellar_mass_halo_mass.yml
```diff
@@ -588,3 +588,152 @@ stellar_mass_halo_mass_M200_centrals_30:
...
+  select_structure_type: 10
+  comment: "Centrals only"
+  legend_loc: "upper left"
+  x:
+    quantity: "masses.mass_200crit"
+    units: Solar_Mass
+    start: 1e13
+    end: 5e15
+  y:
+    quantity: "derived_quantities.projected_stellar_mass_10_kpc"
```
EvgeniiChaikin, line 603: Does the field `derived_quantities.projected_stellar_mass_10_kpc` actually exist? I don't see it defined in the registration file.

The same comment applies to `derived_quantities.projected_stellar_mass` measured in apertures of different size.

colibre/auto_plotter/stellar_mass_halo_mass.yml
```diff
@@ -588,3 +588,152 @@ stellar_mass_halo_mass_M200_centrals_30:
...
+    caption: Includes only central haloes, and excludes the contribution of satellites in projection, for both the simulated and observed data. The halo masses in observations by DeMaio et al. (2020) and the semi-empirical model by Huang et al. (2021) were converted from the $M_{500}$ and $M_\mathrm{BN98}$ definitions, respectively, to a $M_{200}$ definition (critical overdensity), by assuming an NFW profile and concentration $c=5$. The halo masses from Golden-Marx et al. (2022) were converted from a mean overdensity to critical overdensity (factor 200) definition. None of the halo masses are corrected for (hydrostatic and other) bias. The IMF from observations by Golden-Marx et al. (2022) was converted from Salpeter to Chabrier. Cosmology corrections were not needed. The data that are of intermediate-redshift objects can be used for comparison here because the evolution of BCG stellar cores is observed to be negligible in this regime. The Huang et al. (2021) model is tailored to groups and clusters, and uses aperture definitions for stellar masses, not only for its predictions but also for the observational data used for calibration.
```
EvgeniiChaikin, line 620: In the caption, shouldn't it be Huang et al. (2022) instead of Huang et al. (2021)?

The same comment applies to captions of the other figures too.

colibre/auto_plotter/stellar_mass_halo_mass.yml
```diff
@@ -588,3 +588,152 @@ stellar_mass_halo_mass_M200_centrals_30:
...
+    section: Stellar Mass-Halo Mass (BCGs)
+    show_on_webpage: true
+  observational_data:
+    - filename: GalaxyStellarMassHaloMassBCG/Zhang2016_30kpc.hdf5
+    - filename: GalaxyStellarMassHaloMassBCG/Lin2017_30kpc.hdf5
+    - filename: GalaxyStellarMassHaloMassBCG/Kravstov2018_30kpc.hdf5
+    - filename: GalaxyStellarMassHaloMassBCG/GoldenMarx2023_30kpc.hdf5
+    - filename: GalaxyStellarMassHaloMassBCG/Huang2022_30kpc.hdf5
```
EvgeniiChaikin, line 664: Has the submodule `velociraptor-comparison-data` been updated to include the BCG observational data, which was added recently?

EvgeniiChaikin (changes requested):
The PR looks good to me but there are a couple of changes that need to be made before the PR is fully ready. I pointed them out below.

colibre/auto_plotter/black_holes.yml
FHusko: Indeed. I wasn't entirely sure how this works, I thought that the submodule always points to the current master, but that clearly wasn't the case. It should be updated now.

colibre/auto_plotter/stellar_mass_halo_mass.yml
FHusko: I had forgotten to actually add this (critical part), thanks for noticing! These derived projected stellar masses, which I have now added, are just averages over the three axes.

colibre/auto_plotter/stellar_mass_halo_mass.yml
FHusko: Yes, I had been referring to this paper as being published in 2021 when I was starting to add these changes, but then when I found the full reference, the publication date was actually 2022. Clearly I didn't correctly update it everywhere.

colibre/auto_plotter/stellar_mass_halo_mass.yml
FHusko: Should be done now.

EvgeniiChaikin:
Thanks @FHusko! Now everything works as expected!

EvgeniiChaikin (approved):

PR #254, opened 2023-05-21
Add a script that plots the distribution of gas particles' minimal smoothing lengths, split by z

EvgeniiChaikin:
Here is an example of how this will look in the pipeline. The plot on the right is the new one.

Screenshot from 2023-05-21 20-57-02

### DIFF
--- a/colibre/scripts/gas_smoothing_lengths.py
+++ b/colibre/scripts/gas_smoothing_lengths.py
@@ -93,7 +93,6 @@ def make_single_image(filenames, names, h_bounds, number_of_simulations, output_
         ax.axvline(x=h_min, color=line.get_color(), ls="--", lw=0.2)
 
     ax.legend()
-    ax.set_xlim(*h_bounds)
 
     fig.savefig(f"{output_path}/gas_smoothing_lengths.png")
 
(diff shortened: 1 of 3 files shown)


colibre/scripts/gas_smoothing_lengths.py
```diff
@@ -93,7 +93,6 @@ def make_single_image(filenames, names, h_bounds, number_of_simulations, output_
         ax.axvline(x=h_min, color=line.get_color(), ls="--", lw=0.2)
 
     ax.legend()
-    ax.set_xlim(*h_bounds)
```
EvgeniiChaikin, line 96: I removed this so that the lowest limit on h (now set to 1e-4 \times grav soft) is always within the plot's domain.

MatthieuSchaller (approved):

MatthieuSchaller:
The interpretation can be a bit hard I think. You could have a particles with h = 0.001 at z = 2 and then with h = 0.00001 at z=0.1. This then make it disappear from the z=1-3 panel even though it may have been the particle with the lowest h then.

MatthieuSchaller:
It is useful but we'll have to keep that caveat in mind. An alternative is to also plot `min h` vs. `z at which it was reached`.

PR #255, opened 2023-05-24
Add script making a plot with black hole time-steps statistics

EvgeniiChaikin:
Here is an example of running the script on two runs, one at z=2 and the other at z=0. The run in orange uses an outdated code that does not have BH minimal time-bins in the output. The script is compatible with this.

bh_timestep_hist


MatthieuSchaller:
Thanks. I like that better. :)

colibre/scripts/bh_timestep_hist.py
```diff
@@ -0,0 +1,169 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy.interpolate import interp1d
+from matplotlib.lines import Line2D
+from glob import glob
+
+from swiftsimio import load
+import unyt
+
+bin_edges = np.logspace(np.log10(50e-3), np.log10(50e7), 41)  # kyr
```
MatthieuSchaller, line 10: can we document this?

MatthieuSchaller: And the rest of the timebin --> time-step algorithm in case someone else needs to edit this script in the future.

EvgeniiChaikin:
Do you think it it clear enough now? Or should I add anything else?

EvgeniiChaikin:
In the past few days, I was looking at this plot from various runs and came to the conclusion that it is better not to normalise by bin width. This is why I decided to push one last change.

MatthieuSchaller:
Yes, I think the explanation of the conversion is now clear.

MatthieuSchaller (approved):

PR #256, opened 2023-05-26
Add min smoothing length vs. redshift plot and update min smoothing length histrogram plot

EvgeniiChaikin:
Two changes in this update

- Update the gas-particle minimal smoothing-length histogram by normalizing `hmin_gas` by gravitational softening at the corresponding redshift. I.e. before the update, `hmin_gas` would be plotted but now `hmin_gas / \varepsilon_soft` is plotted.

- Add a plot showing gas-particle minimal smoothing-length vs. redshift at which it as reached, plotted as a 2D histogram.

Here are the examples of the updated plot and the new plot
gas_minimal_smoothing_lengths

gas_hmin_redshift

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -223,7 +223,7 @@ scripts:
     section: Histograms
     title: Gas Particle Minimal Smoothing Lengths
   - filename: scripts/gas_minimal_smoothing_length_redshift.py
-    caption: Gas Particle Comoving Minimal Smoothing Lengths versus Redshifts at which these minimal smoothing lengths were reached. The black dashed curve indicates the minimum allowed smoothing length at a given redshift.
+    caption: Gas Particle Comoving Minimal Smoothing Lengths versus Redshifts at which these minimal smoothing lengths were reached. The black dashed and black dotted curves indicate, respectively, the minimum allowed smoothing length and gravitational softening length at a given redshift.
     output_file: gas_hmin_redshift.png
     section: Histograms
     title: Gas Particle Minimal Smoothing Lengths vs. Redshift
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -218,7 +218,7 @@ scripts:
     section: Histograms
     title: Gas Particle Smoothing Lengths
   - filename: scripts/gas_minimal_smoothing_lengths.py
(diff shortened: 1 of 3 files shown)


EvgeniiChaikin:
@MatthieuSchaller, Joop just asked me to also add a curve indicating the redshift evolution of the softening. This is done in the most recent commit.

The new plot will look as follows
gas_hmin_redshift (1)

MatthieuSchaller:
what is e_soft,min ?

EvgeniiChaikin:
The minimum softening (the minimum between the comoving softening and the maximum physical softening).

Since you asked this, I am anticipating that it was not clear, so I should probably just write `e_soft` instead of `e_soft,min`.

MatthieuSchaller:
Is the ratio then exactly 10^-4 ? seems weird to me.

EvgeniiChaikin:
Because in the parameter file, we specify h_min relative to softening.

MatthieuSchaller:
we specify e_plummer in the parameter file.
That means the softening starts at 3 * e_plummer.
We then use H = h_min_ratio *  3 * e_Plummer.
That then leads to h = h_min_ratio * 3 * e_Plummer / kernel_gamma.

EvgeniiChaikin:
Ah, I see what you mean now! Sorry for the confusion!

The script correctly applies the factor of `3` and `kernel_gamma` correction (https://github.com/SWIFTSIM/pipeline-configs/pull/256/files#diff-80007c0d55552991c352811da2c9f718499b849442d3c35e21641247c4c6891fR63) However, the axis does not say it.

What we plot is `smoothing length * kernel_gamma / (3 * e_Plummer)`. This is why the line indicating the `h_min ratio` is exactly at 1e-4.

I purposely did not include `kernel_gamma` and `3` in the label of the X-axis because I didn't want to over-complicate the definition.

Do you think it would be better to change the axis label to something like `h_min / (3 e_Plummer / kernel_gamma),` or just add this extra information to the caption?

MatthieuSchaller:
As long as the calculation is correct then it's ok like this.

EvgeniiChaikin:
I extend the caption. It now says that both the factor of 3 and gamma_kernel are taken into account.

MatthieuSchaller:
Yes, looks good.

MatthieuSchaller (approved):

PR #266, opened 2023-11-08
add konst. data

james-trayford:
Add plots to show the new Konstantopoulou et al 2023 data on DTM and DTG ratios.

These are based on total metallicities rather than O/H values, so also adding versions of the plot demonstrating representative [O/H] values computed from a solar O/H to see how this might affect things.

### DIFF
--- a/colibre/auto_plotter/depletion_relations.yml
+++ b/colibre/auto_plotter/depletion_relations.yml
@@ -36,10 +36,11 @@ oxygen_abundance_v_neutral_dust_to_gas:
     section: Dust Depletion Relations
     show_on_webpage: true
   observational_data:
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_Data_COMW.hdf5
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_BPL_COMW.hdf5
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_Data_COZ.hdf5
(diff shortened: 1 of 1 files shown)


MatthieuSchaller (approved):

colibre/auto_plotter/depletion_relations.yml
```diff
@@ -36,10 +36,11 @@ oxygen_abundance_v_neutral_dust_to_gas:
     section: Dust Depletion Relations
     show_on_webpage: true
   observational_data:
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_Data_COMW.hdf5
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_BPL_COMW.hdf5
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_Data_COZ.hdf5
-    - filename: GalaxyMetallicityDusttoGasRatio/RemyRuyer2014_BPL_COZ.hdf5
+    - filename: GalaxyMetallicityDusttoGasRatio/Konstantopoulou23_z000p250.hdf5
```
EvgeniiChaikin, line 39: We should include the data from Konstantopoulou23 as a single file that contains data points at different redshifts, as opposed to different files for different redshifts.

This will ensure that the data is plotted only at a single redshift per plot (the redshift that is closest to the snapshot's redshift) , as opposed to appearing several times in a given plot.

james-trayford: do you have an example of this being done elsewhere?

PR #272, opened 2024-01-15
Adding arrow in stellar abundance plots

correac:
In this PR we are adding an arrow in the stellar abundance plots to show lower median values [X/Fe] that are not shown in the figure ([Fe/H]<-4). Note that we are also decreasing the lower limit value to -10 in case iron (or other elements) are zero, and we are removing the lower limits of [X/Fe]<-2 that were overwriting some values.

See this figure as example:

stellar_abundances_FeH_MgFe_APOGEE


correac:
@robjmcgibbon could you have a look and merge this change?

PR #273, opened 2024-02-03
Add two new plots with dust depletion relations using a stellar-mass cut

EvgeniiChaikin:
- Add a stellar-mass mask (Mstar > 1e9 Msun) for active galaxies in `registraction.py`
- Add two new plots with dust depletion relations based on the new mask. The simulation data shown in these plots corresponds to galaxies that are not only active but also have Mstar > 1e9 Msun

An example of the dust section with the new plots can be found here: https://swift.dur.ac.uk/COLIBREPlots/chaikin/comparisons/YEAR2023/z0.0_L12N376_vary_dust_clumping/index.html#983660477724058836

An example of the older version of the same section (without the two new plots) can be found here: https://swift.dur.ac.uk/COLIBREPlots/chaikin/comparisons/YEAR2023/z0.0_L12N376_increase_BH_seed_mass/index.html#2422449837951610890

Note that in this PR, I also slightly cleaned  up the `depletion_relations.yml` file, removing the code that is of no use.

### DIFF
--- a/colibre/auto_plotter/depletion_relations.yml
+++ b/colibre/auto_plotter/depletion_relations.yml
@@ -1,46 +1,6 @@
 oxygen_abundance_v_neutral_dust_to_gas:
   type: "scatter"
   legend_loc: "lower right"
-  selection_mask: "derived_quantities.is_active_50_kpc"
-  x:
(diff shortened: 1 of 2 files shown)


robjmcgibbon (approved):
LGTM!

PR #277, opened 2024-03-25
Set aspect ratio for density-temperature plots

robjmcgibbon:
Update the plot aspect ratio so that more ticks will be shown on the x axis. We still end up with only one tick when comparing 7 simulations, but that never really happens anyway. Examples of different numbers of simulations are below.

density_temperature
density_temperature
density_temperature
density_temperature

density_temperature


EvgeniiChaikin:
This is a nice solution! Thanks!!

Would it be possible to also modify the following scripts in the same manner, for consistency?

- bh_masses.py
- density_temperature_species.py
- birth_metallicity_redshift.py
- birth_density_metallicity.py
- birth_density_redshift.py
- birth_density_temperature.py
- particle_updates_step_cost.py
- max_temperature_redshift.py
- density_species_depletions.py
- density_species_diffuse.py
- density_species_interp.py

PR #282, opened 2024-05-01
Add new high-z BH mass - stellar mass data, as well as updated BH mass function data

FHusko:
This accompanies the data added in: https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/196

The plots look like this:

1) BH mass - stellar mass data at redshifts z=1 and z=2
image
image

2) BH mass function
image


robjmcgibbon:
Could you please just update the submodule link? Then this is good to be merged

FHusko:
I've updated it but I am not sure if it's pointing to the correct version of the submodule.

PR #283, opened 2024-06-03
Two new plots with the information on the SN energy fractions

EvgeniiChaikin:
- Add a plot that shows the relation between the SN energy fraction and stellar birth pressure
- Add a plot that shows the relation between the SN energy fraction and galaxy stellar mass
- Rewrite the code in the script that computes the distribution of SN energy fractions, to make it more readable and maintainable

**Example:**

Screenshot from 2024-06-03 16-29-37

The two plots on the right are the new plots.

### DIFF
--- a/colibre/config.yml
+++ b/colibre/config.yml
@@ -103,10 +103,20 @@ scripts:
     section: Stellar Birth Properties
     output_file: birth_pressure_distribution.png
   - filename: scripts/snii_energy_fraction_distribution.py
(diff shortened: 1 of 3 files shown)


robjmcgibbon (approved):
Looks good to me, I was able to run it with no issues. I don't like the fact that we repeat functions so often (e.g. energy_fraction in the PR), but I don't know a nice way around that with the current pipeline

PR #284, opened 2024-06-12
Add separate plots for centrals and satellites

robjmcgibbon:
[Request from Joop](https://leiden-observatory.slack.com/archives/G035RUR984V/p1716973767531199) 
A plot of the central passive fraction was already in the pipeline

Screenshot from 2024-06-07 15-39-25
Screenshot from 2024-06-07 15-39-43
Screenshot from 2024-06-07 15-46-45


robjmcgibbon:
I've merged these changes directly into the `support_new_soap_output` branch since the way we specify structure type in the autoplotter scripts is now different.

PR #296, opened 2025-02-21
Use coupling efficiency when calculating BH luminosity

robjmcgibbon:
For the pipeline we were not considering the BH coupling efficiency when calculating BH luminosities

@FHusko can you please have a quick look


FHusko:
Looks good! Thanks

